PR #43, opened 2022-12-04
chore!: drop ast check for refresh boundary

ArnaudBarre:
The runtime only check is closer to what other implementation are doing including the vite SWC one.

This is mostly for simplification and consistency between plugins. It will also have the benefit of allowing people to know when the fast refresh was skipped when I will port the [invalidate message](https://github.com/vitejs/vite-plugin-react-swc/blob/main/src/index.ts#L100) from the SWC plugin.

cc @IanVS

### DIFF
--- a/packages/plugin-react/src/fast-refresh.ts
+++ b/packages/plugin-react/src/fast-refresh.ts
@@ -1,7 +1,6 @@
 import fs from 'node:fs'
 import path from 'node:path'
 import { createRequire } from 'node:module'
-import type { types as t } from '@babel/core'
 
 export const runtimePublicPath = '/@react-refresh'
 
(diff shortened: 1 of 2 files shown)


IanVS:
I think I'm okay with this, but @aleclarson suggested to keep the AST check in place when I was adding the runtime check.  So maybe he has some thoughts about this.  

Ref: https://discord.com/channels/804011606160703521/1018203348060618782/1019332955551834162

aleclarson:
> maybe he has some thoughts about this.

The AST check avoids an unnecessary runtime check and code transformation in many cases. I haven't measured the performance impact of removing the AST check, so I can't help you decide whether merging this is the best option or not.

patak-cat:
I think it is a good idea to align with other implementations. But let's remove this once we know the perf implications.

ArnaudBarre:
I can do some perf checks but I disagree with:

> The AST check avoids an unnecessary runtime check and code transformation in many cases

The ast check is always done after code transformation, so you can't skip transformation. It's done only when babel already added some "Refrehreg" code, so in >>90% of the cases for codebase that respect rules (and you should otherwise HMR is a nightmare), you will run both AST and runtime checks that both says everything is valid.

aleclarson (approved):

PR #44, opened 2022-12-04
chore: clean some leftovers from Vite core

ArnaudBarre:
The clean up of unused test utils is totally arbitrary. I kept some that would probably be useful in the future like `addFile` but removed some that felt more specific like `formatSourcemapForSnapshot`


ArnaudBarre:
I started a release by mistakes, sure that i wouldn't have npm permissions anyway, but forgot that it would push a git tag to GH.

I deleted the alpha-3 tag, pushed an alpha-2 tag so that the changelog is correct the next time.
 
We need to decide if we do CI publish or not. I see that you updated the TS script but the package.json entry was removed (the CI failure)

patak-cat:
@ArnaudBarre if I understand correctly, you are simplifying the publish script here because there is only one package? I thought we could leave it as is, as we may end up extracting this into a repo utils package later. And we could detect if there is only one option and avoid the prompt only (but the rest remains intact)

ArnaudBarre:
I kept the release script multi-packages enable, just remove `vite` et `create-vite` edge cases that would be always false in the future. I'm pretty sure any published packages in the monorepo will be under `@vitejs` namespace

patak-cat (approved):
Ah, I see, I got confused by some parts of the changelog. Makes sense.

PR #52, opened 2022-12-07
fix: use `addRefreshWrapper` even if only `$RefreshSig$` is used

aleclarson:
### Description

We need to call `addRefreshWrapper` when `$RefreshReg$` **or** `$RefreshSig$` is referenced in a React module. Apparently, there are cases where only `$RefreshSig$` is used, as I just hit one of them. 😃 

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

- [ ] Read the [Contributing Guidelines](https://github.com/vitejs/vite-plugin-react/blob/main/CONTRIBUTING.md).
- [ ] Read the [Pull Request Guidelines](https://github.com/vitejs/vite-plugin-react/blob/main/CONTRIBUTING.md#pull-request-guidelines) and follow the [Commit Convention](https://github.com/vitejs/vite-plugin-react/blob/main/.github/commit-convention.md).
- [ ] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [ ] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -335,7 +335,7 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
 
         if (result) {
           let code = result.code!
-          if (useFastRefresh && /\$RefreshReg\$\(/.test(code)) {
+          if (useFastRefresh && /\$Refresh(Sig|Reg)\$\(/.test(code)) {
             code = addRefreshWrapper(code, id)
           }
           return {


ArnaudBarre (approved):

ArnaudBarre:
I will fix this

Do you have an example of code that generates that? I'm curious.

I need to do the same for the swc plugin

PR #53, opened 2022-12-08
feat!: allow to run Babel on non js/ts extensions (fixes #38)

### LINKED ISSUE #16: Plugin-React: Allow user provided babel plugins to be applied to non-project files
### Clear and concise description of the problem

At present we can provide extra `babel` plugins to the `@vitejs/plugin-react` via its `babel` option, e.g:

```ts
export default defineConfig({
  plugins: [
    react({
      babel: {
        plugins: ["@compiled/babel-plugin"],
      },
    }),
  ],
})
```

For the most part this works fine. The plugins provided are applied to project files, where a project file is not a `node_module` and starts with the `projectRoot` (project root being the current working directory). This is defined [here](https://github.com/vitejs/vite/blob/main/packages/plugin-react/src/index.ts#L113)

This is where I encounter an issue. I am using `pnpm` workspaces in my project and have a directory structure like:

```
root
|
|__vite-demo-app
|          vite.config.ts
|
|__packages/
|         |
|         |__ui-package/
|                    some-files.ts
```

During development my workspace packages use `src/index.ts` for their main entry-point in their `package.json`. This means when a file is loaded it needs to be transpiled (by `vite` in this case). In my case I am using `@compiled/react` which requires its `babel`
plugin to be applied before using the component. Since my packages will be outside the project-root the `react-plugin` will not apply the user provided plugins to it.

### Suggested solution

Allow users to specify an `include` (or something similar) option which can be used to ensure that user provided plugins are still applied. For example, logic defined in the [vite react plugin](https://github.com/vitejs/vite/blob/main/packages/plugin-react/src/index.ts#L116), could become:

```ts
const plugins = isProjectFile || options.include.some(match) ? [...userPlugins] : []
```

Then I could do something like:

```ts
export default defineConfig({
  plugins: [
    react({
      babel: {
        plugins: ["@compiled/babel-plugin"],
        include: [function(filename) { return filename.includes('<my-project-name>/packages/<my-dep>'} ]
      },
    }),
  ],
})
```

The usage I'd ideally like to have seen is similar to how [babel handles situations like this](https://babeljs.io/docs/en/options#include)

### Alternative

In my case I could pre-build my dependencies but this is not a great developer experience. Not aware of any other alternatives - but definitely open to suggestions. 

My main goal in suggesting this change is so that I can make changes to workspace packages and see live reloads without needing to build the workspace package.

### Additional context

Happy to pick this up myself and raise a PR for it.

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitejs/vite/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitejs/vite/blob/main/CONTRIBUTING.md).

### LINKED ISSUE #22: @vitejs/plugin-react can support more extensions
### Clear and concise description of the problem

I'm doing markdown conversion to React component. However, this plugin may only be supported `.jsx?` or `.tsx?` file. So I had to use an extra Babel plugin to handle the files generated by the markdown transformation.

### Suggested solution

Add a configuration extension to the plugin's options and support any text that conforms to React component.

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitejs/vite/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitejs/vite/blob/main/CONTRIBUTING.md).

### LINKED ISSUE #24: `@vitejs/plugin-react` processing excluded files
### Describe the bug

`@vitejs/plugin-react` processing files that defined in `exclude` option.

```ts
export default defineConfig({
  plugins: [
    react({
      exclude: /App/,
    }),
  ],
});
```

I was expecting `@vitejs/plugin-react` to bypass processing matching files and throw an error but it's still processing those files.

`include` and `exclude` filters are only checked when loading react-refresh plugin and jsx transforms are still processes for all jsx files:
https://github.com/vitejs/vite/blob/e8c840abd2767445a5e49bab6540a66b941d7239/packages/plugin-react/src/index.ts#L170

In my opinion we should do that on top of transform function by adding this:
```ts
  if (!filter(id)) {
    return
  }
  // ...  
```

#### **Why I do this?**
I'm currently using Solid and React on same project and both of them have their own jsx transform. I need to prevent them processing the same file. I have naming convention in my app `.react.tsx` for React components and `.solid.tsx` for Solid components. I didn't include Solid in my repro because vite-plugin-solid [doesn't have an option to exclude files](https://github.com/solidjs/vite-plugin-solid/blob/b57f3e9ed8c3048afe7d1f33bb85b4daefad2e03/src/index.ts#L308) but I patched their plugin locally and will make a feature request on their repo. 

### Reproduction

https://stackblitz.com/edit/vitejs-vite-k286cf?file=vite.config.js&terminal=dev

### System Info

```shell
System:
    OS: Windows 10 10.0.19042
  Binaries:
    Node: 16.13.1 - C:\Program Files\nodejs\node.EXE
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitejs/vite/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitejs/vite/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/vitejs/vite/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a Vite issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/core instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitejs/vite/discussions) or join our [Discord Chat Server](https://chat.vitejs.dev/).

### LINKED ISSUE #38: Need a way to transform other plugin(such as mdx)'s output.
### Description

I am currently using mdx with react. I use [@mdx-js/rollup](https://github.com/brillout/vite-plugin-mdx/issues/44#issuecomment-974540152) transform .mdx files into React components.
But neither @mdx-js/rollup or @vitejs/plugin-react is doing react-refresh transform to the .mdx files (.i.e .mdx files are not [self-accepting](https://vitejs.dev/guide/api-hmr.html#hot-accept-cb)). Editing .mdx file during dev results in full reload if it not imported by a hmr boundary (.e.g imported by a .jsx file which is made self-accepting by @vitejs/plugin-react).
> @vitejs/plugin-react options like `exclude` and `include` don't work.

Here is a example:
https://stackblitz.com/edit/vitejs-vite-ucnnpy?file=vite.config.ts,src%2Fdemo.mdx,src%2Fmain.tsx,src%2FWrapper.tsx,src%2Fvite-env.d.ts&terminal=dev

Can @vitejs/plugin-react provide a way to transform @mdx-js/rollup’s output?

### Suggested solution

@vitejs/plugin-react should be able to handle other plugin's output.

This line should be adjustable by option so that `.mdx` can pass this check
https://github.com/vitejs/vite-plugin-react/blob/b647e74c38565696bd6fb931b8bd9ac7f3bebe88/packages/plugin-react/src/index.ts#L215

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [x] Follow our [Code of Conduct](https://github.com/vitejs/vite/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitejs/vite/blob/main/CONTRIBUTING.md).

ArnaudBarre:
This PR is currently composed of 3 commits:

- First, just skip everything if id include node_modules. Given that esbuild pre-bundle deps, it seems strange for me that you need to process them again via Babel. I'm not sure we should encourage people to rely on the fact that they use node_modules with non standard content
- Second, main commit: use the include/exclude like other plugins, with a default include that was the previous mandatory check. I don't see why you would want to skip fast refresh if it can be enable. This simplifies a lot the code and the configuration options compare to https://github.com/vitejs/vite/pull/6202. Finally I think that changing the behaviour when the import is outside vite root but still imported as source is confusing. I have bad memories of CRA disabling simple code sharing between two projects and requiring to setup a fake local lib. 
- Third, add an mdx playground with a simple HMR test

Fixes #38, fixes #22, fixes #24 and should also fix #16


patak-cat:
I think we should wait for this one a bit, given that the release of Vite 4 is so close. Now that the plugin is is in a separate repository we can do a new major for it independent of Vite after the holidays. We can do the a proper beta period and check with the ecosystem.

packages/plugin-react/src/index.ts
```diff
@@ -211,11 +216,8 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
         filepath.match(fileExtensionRE) ||
         []
 
-      if (/\.(?:mjs|[tj]sx?)$/.test(extension)) {
+      if (filter(id)) {
```
sapphi-red, line 219: I think this changes behavior when `id` is `foo.txt?a=.jsx`. (it was `true` but now it's `false`)

PR #122, opened 2023-03-26
feat!: allow to run Babel on non js/ts extensions

### LINKED ISSUE #16: Plugin-React: Allow user provided babel plugins to be applied to non-project files

### LINKED ISSUE #22: @vitejs/plugin-react can support more extensions

### LINKED ISSUE #24: `@vitejs/plugin-react` processing excluded files

### LINKED ISSUE #38: Need a way to transform other plugin(such as mdx)'s output.

### LINKED ISSUE #83: Fast Refresh not working with automatic runtime and file extension without "x"
### Describe the bug

Currently, fast refresh is working for a file only if either

- the file extension ends with an "x" (e.g. "jsx", "tsx") *or*
- the code in the file imports `React` in some way

Now when using a compile-to-JS language like [ReScript](https://rescript-lang.org/) **with the new JSX runtime enabled**, one can get `.js` or `.mjs` files with imports from "react/jsx-runtime" or "react/jsx-dev-runtime", but no `React` import, and fast refresh will not work for these files.

### Reproduction

https://github.com/cknitt/vite-plugin-react-fast-refresh-issue

### Steps to reproduce

1. Install dependencies and start vite:
```sh
npm i
npm run dev
```
2. Increment counter
3. Change the text in `TestComponent.bs.js` (alternatively change it in TestComponent.res and run `npm run res:build` to recompile). Observe that the counter value is reset to 0.

### Workaround

1. Run `npx patch-package` to apply the patch from `patches/@vitejs+plugin-react+3.0.1.patch`.
2. Restart vite and repeat the above test steps.
3. Observe that TestComponent now refreshs correctly and the counter value remains unchanged.

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 893.44 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.17.1 - ~/Library/Caches/fnm_multishells/60384_1673351682301/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.15.0 - ~/Library/Caches/fnm_multishells/60384_1673351682301/bin/npm
    Watchman: 2023.01.02.00 - /opt/homebrew/bin/watchman
  Browsers:
    Brave Browser: 108.1.46.153
    Firefox: 107.0.1
    Safari: 16.2
  npmPackages:
    @vitejs/plugin-react: ^3.0.1 => 3.0.1
    vite: ^4.0.4 => 4.0.4
```

### Used Package Manager

npm

### Logs

_No response_

### Validations

ArnaudBarre:
- Superseed #53
- Fixes #38
- Fixes #22
- Fixes #24
- Fixes #16
- Fixes #83

API changes:
- Make include/exclude work like other plugins. This allows to make it work with mdx plugin. Also I think that all the query params thing was to workaround this limitation (and probably also inspire by how Vue SFC works) and I don't think it makes sense for the React plugin now that you can explicitly include more extensions.
- Don't change the behaviour when the import is outside of Vite root. This is blocking a lot of monorepo setup.
- Don't process node_modules files. TS in node_modules is still supported thanks to esbuild. It would be really strange to require a custom babel transform to load a node_modules file.
- Drop fastRefresh: I don't see any reason for this flag. This is 80% of the goal of this plugin now that esbuild handles all jsx transformation. If there are other cases where it needs to be disabled (like ssr, prod currently), this should be detected by the plugin.


packages/plugin-react/src/index.ts
```diff
@@ -16,27 +15,16 @@ export interface Options {
   include?: string | RegExp | Array<string | RegExp>
   exclude?: string | RegExp | Array<string | RegExp>
   /**
-   * Enable `react-refresh` integration. Vite disables this in prod env or build mode.
-   * @default true
-   */
-  fastRefresh?: boolean
-  /**
-   * Set this to `"automatic"` to use [vite-react-jsx](https://github.com/alloc/vite-react-jsx).
+   * @deprecated
```
patak-cat: I think it would be good to add a comment here about why the option is deprecated, and also say it will be removed in the next Major.

packages/plugin-react/src/index.ts
```diff
@@ -89,261 +77,198 @@ declare module 'vite' {
...
   let devBase = '/'
-  let filter = createFilter(opts.include, opts.exclude)
+  const filter = createFilter(
+    [
+      defaultIncludeRE,
+      ...(Array.isArray(opts.include)
+        ? opts.include
+        : opts.include
+        ? [opts.include]
+        : []),
```
patak-cat: Maybe you could copy the helper we use in Vite core so this is easier to follow for others
```
export function arraify<T>(target: T | T[]): T[] {
  return Array.isArray(target) ? target : [target]
}
```

packages/plugin-react/src/index.ts
```diff
@@ -378,59 +312,7 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
...
-          `export const jsxs = jsxRuntime.jsxs`,
-        ].join('\n')
-      }
-      if (id === virtualReactJsxDevRuntimeId) {
-        return [
-          `import * as jsxRuntime from ${JSON.stringify(reactJsxDevRuntimeId)}`,
-          `export const Fragment = jsxRuntime.Fragment`,
-          `export const jsxDEV = jsxRuntime.jsxDEV`,
-        ].join('\n')
-      }
```
patak-cat: I think the idea here was to do it manually to improve the output code. And if you don't have this plugin, you'll see a lot of `__vite_xxx` names instead. @aleclarson may know better why this was needed, and I think @bluwy also was touching this plugin at one point.

ArnaudBarre: Oh ok I get the point, I will take a closer look at this

ArnaudBarre: I just looked at the output, feels ok for the jsx naming. The import line is a bit long but not sure anybody look at this.
And this would be fixed when React _finally_ ships esm.
```js
import { injectQuery as __vite__injectQuery } from "/@vite/client";import { createHotContext as __vite__createHotContext } from "/@vite/client";import.meta.hot = __vite__createHotContext("/App.jsx");import __vite__cjsImport0_react_jsxDevRuntime from "/node_modules/.vite/deps/react_jsx-dev-runtime.js?v=88111a26"; const jsxDEV = __vite__cjsImport0_react_jsxDevRuntime["jsxDEV"];
import RefreshRuntime from "/@react-refresh";
let prevRefreshReg;
let prevRefreshSig;
if (import.meta.hot) {
  if (!window.__vite_plugin_react_preamble_installed__) {
    throw new Error("@vitejs/plugin-react can't detect preamble. Something is wrong. See https://github.com/vitejs/vite-plugin-react/pull/11#discussion_r430879201");
  }
  prevRefreshReg = window.$RefreshReg$;
  prevRefreshSig = window.$RefreshSig$;
  window.$RefreshReg$ = (type, id) => {
    RefreshRuntime.register(type, "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx " + id);
  };
  window.$RefreshSig$ = RefreshRuntime.createSignatureFunctionForTransform;
}
var _s = $RefreshSig$();
import __vite__cjsImport3_react from "/node_modules/.vite/deps/react.js?v=b9578cd9"; const useState = __vite__cjsImport3_react["useState"];
import Button from "/@fs/Users/arnaud/git/vite-plugin-react/node_modules/.pnpm/file+playground+react+jsx-entry/node_modules/jsx-entry/Button.jsx";
import Dummy from "/components/Dummy.jsx?qs-should-not-break-plugin-react";
import Parent from "/hmr/parent.jsx";
import { CountProvider } from "/context/CountProvider.jsx";
import { ContextButton } from "/context/ContextButton.jsx";
function App() {
  _s();
  const [count, setCount] = useState(0);
  return /* @__PURE__ */ jsxDEV("div", { className: "App", children: [
    /* @__PURE__ */ jsxDEV("header", { className: "App-header", children: [
      /* @__PURE__ */ jsxDEV("h1", { children: "Hello Vite + React" }, void 0, false, {
        fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
        lineNumber: 13,
        columnNumber: 9
      }, this),
      /* @__PURE__ */ jsxDEV("p", { children: /* @__PURE__ */ jsxDEV("button", { id: "state-button", onClick: () => setCount((count2) => count2 + 1), children: [
        "count is: ",
        count
      ] }, void 0, true, {
        fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
        lineNumber: 15,
        columnNumber: 11
      }, this) }, void 0, false, {
        fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
        lineNumber: 14,
        columnNumber: 9
      }, this),
      /* @__PURE__ */ jsxDEV("p", { children: /* @__PURE__ */ jsxDEV(ContextButton, {}, void 0, false, {
        fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
        lineNumber: 20,
        columnNumber: 11
      }, this) }, void 0, false, {
        fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
        lineNumber: 19,
        columnNumber: 9
      }, this),
      /* @__PURE__ */ jsxDEV("p", { children: [
        "Edit ",
        /* @__PURE__ */ jsxDEV("code", { children: "App.jsx" }, void 0, false, {
          fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
          lineNumber: 23,
          columnNumber: 16
        }, this),
        " and save to test HMR updates."
      ] }, void 0, true, {
        fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
        lineNumber: 22,
        columnNumber: 9
      }, this),
      /* @__PURE__ */ jsxDEV("a", { className: "App-link", href: "https://reactjs.org", target: "_blank", rel: "noopener noreferrer", children: "Learn React" }, void 0, false, {
        fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
        lineNumber: 25,
        columnNumber: 9
      }, this)
    ] }, void 0, true, {
      fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
      lineNumber: 12,
      columnNumber: 7
    }, this),
    /* @__PURE__ */ jsxDEV(Dummy, {}, void 0, false, {
      fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
      lineNumber: 30,
      columnNumber: 7
    }, this),
    /* @__PURE__ */ jsxDEV(Parent, {}, void 0, false, {
      fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
      lineNumber: 31,
      columnNumber: 7
    }, this),
    /* @__PURE__ */ jsxDEV(Button, { children: "button" }, void 0, false, {
      fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
      lineNumber: 32,
      columnNumber: 7
    }, this)
  ] }, void 0, true, {
    fileName: "/Users/arnaud/git/vite-plugin-react/playground/react/App.jsx",
    lineNumber: 11,
    columnNumber: 10
  }, this);
}
_s(App, "oDgYfYHkD9Wkv4hrAPCkI/ev3YU=");
_c = App;
```

ArnaudBarre:
Update: remove the optimizeDeps include. This avoid forcing people using another jsxRuntime to pass it just for that and Vite is able to pre-bundle it on the first run in my playground testing

packages/plugin-react/src/index.ts
patak-cat: Interesting, yes, this looks fine to me too 🤔
Maybe there was another issue trying to be solved, but if the plugin works without this extra complexity it is worth trying it in the major beta, and you could revert later. Maybe it isn't bad to break this PR into a few parts (removing this being one for example) so it is easier to reference later.

aleclarson: It was originally added for more minifiable `vite build` output (using `jsx(...)` instead of something like `jsxRuntime.jsx(...)`).

ArnaudBarre: Just tested and the build output is ok IMO:
`F.jsxs("p",{children:["Edit ",F.jsx("code",{children:"App.jsx"})," and save to test HMR updates."]})`

aleclarson: `F.jsxs` vs `F` is an 83% difference, which is pretty significant if you have a large app. Even though network compression removes the issue for delivery speed, there's still parsing performance. That said, I haven't benchmarked the difference in a large app, so it's only conjecture from my end.

ArnaudBarre: yeah gzip size difference should be really small. The extra parsing time/execution time seems like a bit over optimisation for me. I just hope the React team ships esm before 2030

packages/plugin-react/src/index.ts
```diff
@@ -83,32 +78,32 @@ declare module 'vite' {
...
 export default function viteReact(opts: Options = {}): PluginOption[] {
   // Provide default values for Rollup compat.
   let devBase = '/'
-  let filter = createFilter(opts.include, opts.exclude)
+  const filter = createFilter(
+    opts.include === undefined
+      ? defaultIncludeRE
+      : [defaultIncludeRE, ...arraify(opts.include)],
+    opts.exclude,
+  )
```
sapphi-red: > I'm still unsure if the default include should be override or not

I prefer overriding because other plugins does that.

- plugin-vue: https://github.com/vitejs/vite-plugin-vue/blob/328fc10840a63e8b8e1e2e8286b6cfd978202ef8/packages/plugin-vue/src/index.ts#L85
- plugin-vue-jsx: https://github.com/vitejs/vite-plugin-vue/blob/328fc10840a63e8b8e1e2e8286b6cfd978202ef8/packages/plugin-vue-jsx/src/index.ts#L42
- vite-plugin-svgr: https://github.com/pd4d10/vite-plugin-svgr/blob/ee1bffd101dad54837f5493882824a764c80c3b6/src/index.ts#L25
- vite-plugin-dts: https://github.com/qmhc/vite-plugin-dts/blob/d3843976fcf861ebab8dd9ff373f635957f034df/src/plugin.ts#L276

packages/plugin-react/README.md
```diff
@@ -17,32 +17,33 @@ export default defineConfig({
...
 ```
 
-### Configure the JSX import source
+### jsxImportSource
 
 Control where the JSX factory is imported from. For TS projects this is inferred from the tsconfig.
 
 ```js
 react({ jsxImportSource: '@emotion/react' })
 ```
```
sapphi-red: Do we need this option? We have `esbuild.jsxImportSource` option in the Vite's config.
I wonder if we can drop this one too like we'll do with `jsxPure` option.

packages/plugin-react/src/index.ts
```diff
@@ -167,145 +158,117 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
       }
     },
     async transform(code, id, options) {
+      if (id.includes('/node_modules/')) return
```
sapphi-red, line 154: I think it would be good to write this behavior (skipping `node_modules`) in the readme.

packages/plugin-react/src/index.ts
```diff
@@ -167,145 +158,117 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
...
+        return newBabelOptions
+      })()
+      const plugins = [...babelOptions.plugins]
+
+      const useFastRefresh = !skipFastRefresh && !ssr
+      if (useFastRefresh) {
+        plugins.push([
+          await loadPlugin('react-refresh/babel'),
+          { skipEnvCheck: true },
+        ])
```
sapphi-red, line 186: This line seems to run every file. Before this PR, this part was inside `if ((isJSX || importReactRE.test(code)) && filter(id))`.
I guess this change is for https://github.com/vitejs/vite-plugin-react/issues/83.

Previously we were able to skip babel completely for a non-react file. I'm not sure how much this will impact the performance but I guess this may make a normal react project slower.

ArnaudBarre:
I think in the next PR I will remove the "pre" for the transform part (but keep it for the resolve refresh runtime). WDY @patak-dev?

packages/plugin-react/src/index.ts
```diff
@@ -167,145 +158,117 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
...
+      const parserPlugins: typeof babelOptions.parserOpts.plugins = [
+        ...babelOptions.parserOpts.plugins,
+        'importMeta',
+        // This plugin is applied before esbuild transforms the code,
+        // so we need to enable some stage 3 syntax that is supported in
+        // TypeScript and some environments already.
+        'topLevelAwait',
+        'classProperties',
+        'classPrivateProperties',
+        'classPrivateMethods',
```
sapphi-red: I guess this is out of scope of this PR but leaving a comment as a note.

---

This plugin now depends on babel 7.21.3.
https://github.com/vitejs/vite-plugin-react/blob/e26bb6747f913d243f474c2bf2435f49cafb6b1b/packages/plugin-react/package.json#L42
babel 7.21.3 enables `topLevelAwait`, `classProperties`, `classPrivateProperties`, `classPrivateMethods` by default. So we can remove this part.
https://babeljs.io/docs/babel-parser#latest-ecmascript-features

I'm not sure about `importMeta`.

packages/plugin-react/README.md
ArnaudBarre: As this is inferred for TS project we could go for that. This is just less simple/intuitive for JS projects

packages/plugin-react/src/index.ts
ArnaudBarre: I think this is an expected behaviour?
I can mention it in the changelog, but that strange for me that you will apply some very custom babel preset on npm libs

packages/plugin-react/src/index.ts
ArnaudBarre: Good point I will do a cleanup in a next PR

packages/plugin-react/src/index.ts
ArnaudBarre: Oh yes that's a good point.

Maybe we should go one step further and default to only applying the plugin is JSX/TSX file people can use the include to add more to it.

packages/plugin-react/src/index.ts
aleclarson: Doesn't this break `babel.config.js`?

> You want to compile node_modules?
>
>    [babel.config.json](https://babeljs.io/docs/configuration#babelconfigjson) is for you!

https://babeljs.io/docs/configuration

packages/plugin-react/src/index.ts
aleclarson: I don't see why we wouldn't keep doing it the same way (ie: using regex to check for `react` import in JS files).

packages/plugin-react/src/index.ts
ArnaudBarre: esbuild still runs on node_module so I think this is fine. I would really surprised that people need to run a trasnformation not supported by esbuild on node_modules.

sapphi-red: > I think this is an expected behaviour?

I agree with this, but I think it'd be good that people will be aware of this.

packages/plugin-react/README.md
ArnaudBarre: I'm keeping it because this can the best check I can think of for detecting jsx inside non jsx files.

sapphi-red (approved):

PR #127, opened 2023-03-29
chore: drop interop of jsxRuntime

ArnaudBarre:
Re: https://github.com/vitejs/vite-plugin-react/pull/122#pullrequestreview-1362582761


patak-cat (approved):
I think it makes sense to try this. I would like @bluwy's opinion too here.

bluwy (approved):
LGTM. Good to remove complexity if it doesn't improve a lot.

PR #129, opened 2023-03-29
chore!: drop jsxPure

ArnaudBarre:
This is a really niche custom property that is passed down to esbuild. If you need it, set jsxSideEffects: true in the esbuild config directly.


patak-cat (approved):
Given how small usage this has, this could be done in this major. I think it may not be a bad idea to add a message if `jsxPure` is defined in the config though to point people to use `jsxSideEffects` (at least having the message during the next major).

patak-cat (approved):

PR #130, opened 2023-03-29
chore: refactor runPluginOverrides

ArnaudBarre:
The nested callback and the boolean returns confused me.
I prefer to have the run callback optional and use an iife for setup the babelPlugins so I this doesn't need a temporary let and there is one less branch

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -99,10 +99,9 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
   let projectRoot = process.cwd()
   let skipFastRefresh = opts.fastRefresh === false
   const skipReactImport = false
-  let runPluginOverrides = (
-    options: ReactBabelOptions,
-    context: ReactBabelHookContext,
-  ) => false
+  let runPluginOverrides:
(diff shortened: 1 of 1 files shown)


packages/plugin-react/src/index.ts
```diff
@@ -218,17 +214,16 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
...
-          if (!runPluginOverrides(babelOptions, { ssr, id: id })) {
-            staticBabelOptions = babelOptions
-          }
-        }
+        const babelOptions = (() => {
+          if (staticBabelOptions) return staticBabelOptions
+          const newBabelOptions = createBabelOptions(
+            typeof opts.babel === 'function'
+              ? opts.babel(id, { ssr })
+              : opts.babel,
```
patak-cat, line 192: We lost an optimization here, if `opts.babel` isn't a function, `createBabelOptions` can be created upfront and cached.

ArnaudBarre: I kept it when there is not plugin overrides.
When there are one, I think it's safer to re-create one because it mutates the object

patak-cat: Oh, I see 👍🏼

patak-cat (approved):

PR #167, opened 2023-05-31
fix(plugin-react): import react-refresh babel with full path

augustobmoura:
### Description

In a few cases where the plugin is being imported from a ESM MJS environment the import for `react-refresh` fails. The reason for that is that ESM NodeJS doesn't accept bare extensions imports, the full path is needed.

Because this package distributes a `.mjs` version of the plugin, we should comply with this rule from ESM nodejs 

Specifying the full path with extensions shouldn't be a problem with CJS, and only impact codes that is already setup as MJS.

### Additional context

One such error was found by me while migrating a project from Storybook v6 webpack to Storybook v7 Vite. I don't have the full reason why the `.mjs` distributed file was pick up, but manually patching the import to `react-refresh/babel.js` fixes the issue.

Screenshot of the Storybook error:
image

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -174,7 +174,7 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
             : importReactRE.test(code)))
       if (useFastRefresh) {
         plugins.push([
-          await loadPlugin('react-refresh/babel'),
+          await loadPlugin('react-refresh/babel.js'),
           { skipEnvCheck: true },
         ])
       }


ArnaudBarre:
Something is breaking in the test suite. I will debug this tomorrow

augustobmoura:
Hmm, it looks like react-refresh already [re-exported babel as babel.js](https://github.com/vitejs/vite-plugin-react/actions/runs/5128141170/jobs/9225274366?pr=167#step:11:118), the problem I had might be related to a diff thing. Let me check again on my project and I come back to this PR

augustobmoura:
Found the issue locally, it was a weird resolution problem with yarn, react-refresh should work on the latest version

PR #170, opened 2023-06-03
feat: vitest deps.experimentalOptimizer

ArnaudBarre:
So actually this change is less safer than I though because it can break people using testing Vite HMR like @brillout.

I will merge this now and see the result in ecosystem CI tomorrow. My current idea is that this is rare and asking few framework authors to set this variable is ok.

Another more long term fix could be for vitest to add a flag in the transform call (like `ssr`) so that plugin can adapt more properly to this case.

Related PR in the [SWC plugin](https://github.com/vitejs/vite-plugin-react-swc/pull/115). Original issue: https://github.com/vitejs/vite-plugin-react-swc/issues/111

### DIFF
--- a/packages/plugin-react/CHANGELOG.md
+++ b/packages/plugin-react/CHANGELOG.md
@@ -1,4 +1,8 @@
+# Changelog
+
 ## Unreleased
+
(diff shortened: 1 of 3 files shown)


ArnaudBarre:
@patak-dev So this is failing. Should we continue in this direction or find another way?

PR #188, opened 2023-06-27
fix: fix fast-refresh for files that are transformed into jsx

csr632:
### Description

Previous [useFastRefresh check](https://github.com/vitejs/vite-plugin-react/blob/f8bfc8d45b5a815ae94f9934018b740e5f5cf4e2/packages/plugin-react/src/index.ts#L176) is wrong. It causes fast-refresh not working for files with extension such as `.md`. Checkout the updated playground and the test as a demo.

### Additional context

The first commit update the test that fails and demonstrates the issue. The second commit fix the issue and make the test pass.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -117,9 +117,9 @@ To use breakpoints and explore code execution, you can use the ["Run and Debug"]
 
 Some errors are masked and hidden away because of the layers of abstraction and sandboxed nature added by Vitest, Playwright, and Chromium. In order to see what's actually going wrong and the contents of the devtools console in those instances, follow this setup:
 
-1. Add a `debugger` statement to the `playground/vitestSetup.ts` -> `afterAll` hook. This will pause execution before the tests quit and the Playwright browser instance exits.
+1. Add a `debugger` statement to the `playground/vitestSetup.ts` -> `afterAll` hook. This will pause execution before the tests quit and the Playwright browser instance exits. You can also add `debugger` statements in your tests (`.spec.ts`) files.
 
-2. Run the tests with the `debug-serve` script command, which will enable remote debugging: `pnpm run debug-serve react-sourcemap`.
+2. Run the tests with the `debug-serve` script command, which will enable remote debugging: `pnpm run debug-serve react-sourcemap`. Remember to run this command within the **"JavaScript Debug Terminal"** of VS Code.
 
 3. Wait for inspector devtools to open in your browser and the debugger to attach.
 
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -173,8 +173,8 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
         !ssr &&
         (isJSX ||
           (opts.jsxRuntime === 'classic'
-            ? code.includes(devRuntime)
-            : importReactRE.test(code)))
+            ? importReactRE.test(code)
+            : code.includes(devRuntime)))
       if (useFastRefresh) {
         plugins.push([
           await loadPlugin('react-refresh/babel'),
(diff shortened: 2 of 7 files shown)


ArnaudBarre:
Oh yeah this check is clearly a refactor issue.

What's the use case for having JSX inside `.md`?

csr632:
> What's the use case for having JSX inside .md?

`.md` is transformed into a jsx syntax, and is expected to have fast-refresh added by `vite-plugin-react` (so that editing `.md` files won't trigger full page reload). But the [current useFastRefresh check](https://github.com/vitejs/vite-plugin-react/blob/f8bfc8d45b5a815ae94f9934018b740e5f5cf4e2/packages/plugin-react/src/index.ts#L176) result is false and `vite-plugin-react` is not transforming it. (So currently editing `.md` files is triggering full page reload)

Currently `.mdx` is working because [isJSX](https://github.com/vitejs/vite-plugin-react/blob/f8bfc8d45b5a815ae94f9934018b740e5f5cf4e2/packages/plugin-react/src/index.ts#L170C3-L170C3)(which is `filepath.endsWith('x')`) is true. But it is not working for`.md`.

> mdx.js support both `md` format and `mdx` format: https://mdxjs.com/packages/mdx/#optionsformat

ArnaudBarre (approved):

jprosevear:
Possibly related - I have a project ported over from create-react-app (and which used JS files).  HMR will not work on .js files until I do one of the following:

1) Add import React from 'react'; at the top of the file

2) Change the file extension to .jsx

csr632:
> on .js files

Vite emits error if you use jsx syntax in `.js` file. Why ary you not getting error when doing this? Are you using extra babel rules to do the jsx transform? 

Example: https://stackblitz.com/edit/vitejs-vite-cyjzmi?file=src%2FApp.js

Can you reproduce your issue in this demo?

jprosevear:
```
    // https://github.com/vitejs/vite/discussions/3448#discussioncomment-749919
    esbuild: {
      loader: 'jsx',
      include: /src\/.*\.jsx?$/,
      // loader: "tsx",
      // include: /src\/.*\.[tj]sx?$/,
      exclude: []
    },
    optimizeDeps: {
      esbuildOptions: {
        plugins: [
          {
            name: 'load-js-files-as-jsx',
            setup(build) {
              build.onLoad({ filter: /src\/.*\.js$/ }, async (args) => ({
                loader: 'jsx',
                contents: await fs.readFile(args.path, 'utf8')
              }));
            }
          }
        ]
      }
    }
```

jprosevear:
https://stackblitz.com/edit/vitejs-vite-p5o1nn?file=src%2FApp.js

image

(Changing something in App.js)

csr632:
After some digging, I think the issue is similar ([incorrect useFastRefresh](https://github.com/csr632/vite-plugin-react-1/commit/50d664caca105192ee499a3b603d54edff8c2fba)) but it is not fixed by this PR. I have added a reproduction to the plugin repo: https://github.com/csr632/vite-plugin-react-1/tree/react-js-ext-demo.
So the `useFastRefresh` judgement is still very fragile even after this PR. It can be improve, but **should we improve it for this special case?** As the discuession https://github.com/vitejs/vite/discussions/3448#discussioncomment-749919 said, using jsx syntax in `.js` file is not recommended. I would recommend you to migrate `.js` to `.jsx` instead: https://github.com/vitejs/vite/discussions/3448#discussioncomment-3183056

jprosevear:
Its not unreasonable to js -> jsx, I was just trying to minimize the number of changes in order to do the initial port.  It may also be a common case converting from create-react-app which may help uptake.

jprosevear:
> After some digging, I think the issue is similar ([incorrect useFastRefresh](https://github.com/csr632/vite-plugin-react-1/commit/50d664caca105192ee499a3b603d54edff8c2fba)) but it is not fixed by this PR.

I tried locally installing the package from a local repo checkout with this change.  As you note, didn't work.

PR #224, opened 2023-09-23
feat: add support for non-dev jsx runtime

jihchi:
### Description

I stumble across an issue where the *Fast Refresh* feature doesn't work with [a Vite plugin](https://github.com/jihchi/vite-plugin-rescript) I'm maintaining. After examination, I identified the root cause: The generated code with `.js` extension uses `react/jsx-runtime` for JSX transformation instead of `react/jsx-dev-runtime`, because of this, the generated code is no longer eligible for applying Fest Refresh, which results in a full page reloaded.

<details> <summary>Here is a JS file generated by the compiler, for example (click to expand)</summary>

Filename: `Res.bs.js`

```js
// Generated by ReScript, PLEASE EDIT WITH CARE

import * as JsxRuntime from "react/jsx-runtime";

function Res(props) {
  return JsxRuntime.jsx("h3", {
    children: "ReScript Component",
    className: "text-blue-100",
  });
}

var make = Res;

export { make };
/* react/jsx-runtime Not a pure module */
```
</details>

In order to fix it, this PR adds an extra check for such case so that the Fast Refresh works as expected.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -98,7 +98,9 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
   // Provide default values for Rollup compat.
   let devBase = '/'
   const filter = createFilter(opts.include ?? defaultIncludeRE, opts.exclude)
-  const devRuntime = `${opts.jsxImportSource ?? 'react'}/jsx-dev-runtime`
+  const jsxImportSource = opts.jsxImportSource ?? 'react'
+  const jsxImportRuntime = `${jsxImportSource}/jsx-runtime`
+  const jsxImportDevRuntime = `${jsxImportSource}/jsx-dev-runtime`
   let isProduction = true
   let projectRoot = process.cwd()
   let skipFastRefresh = false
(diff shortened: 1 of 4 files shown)


ArnaudBarre:
Is there any reason why ReScript is not generating code with jsxDev? This could have some benefits like adding information the for react devtools and making it work with https://github.com/ArnaudBarre/vite-plugin-react-click-to-component

jihchi:
Hi @ArnaudBarre, I have no idea. I've done some searching but there's no insightful result. For what it's worth, it [appears](https://github.com/rescript-lang/rescript-react/pull/49/files#diff-cf522c7512e1ca6617a05e4db2dc3a6a4594a7fc71ae2862177a1d146a86c8edR39) that it was implemented this way initially.

As far as I'm concerned, it is likely a challenge and substantial effort to make jsxDev works seamlessly in the compiler.

ArnaudBarre:
Thanks for the great PR. I will merge this as there is also other cases where this can show up https://github.com/vitejs/vite-plugin-react-pages/issues/132

PR #242, opened 2023-10-21
fix: file extension could be mocked/overridden in querystring

c0dedance:
### Description

The file extension can be mocked or overridden using the query string in the fragment identifier ("#").
When the import path contains "#jsx", it indicates that the module will be treated as a JSX module.

image

> When importing modules, some common query symbols that can be used without affecting the import itself include:
?: Used to add a query string parameter. For example: import module from './module.tsx?key=value'.
#: Used to add a fragment identifier, which specifies a specific fragment when importing. This is often used when importing CSS stylesheets. For example: import './styles.css#section1'.
### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -167,7 +167,7 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
     async transform(code, id, options) {
       if (id.includes('/node_modules/')) return
 
-      const [filepath] = id.split('?')
+      const [filepath] = id.split(/\?|#/)
       if (!filter(filepath)) return
 
       const ssr = options?.ssr === true


ArnaudBarre:
Where does this quote come from? I can't find it on Google
What is the usecase for this?

c0dedance:
> Where does this quote come from? I can't find it on Google What is the usecase for this?

When I was writing a plugin, I wanted to inject react-refresh-related code into .mdx files as well. So, I used the transform function to manually inject it.

plugin.transform?.call(this, code, id + '?.jsx', options) // Trick it to handle non-jsx/tsx files for me, but the new version of the plugin has fixed this case.

plugin.transform?.call(this, code, id + '#.jsx', options) // A new way to bypass the restriction.

The information provided in the resource shared by [GPT](https://poe.com/s/epRIacZssbUkRqTjyq3t), which I was told, may be correct.

ArnaudBarre:
This hack was drop in the last major in favor of a more clean include option.
You can check this example: https://github.com/vitejs/vite-plugin-react/blob/main/playground/mdx/vite.config.ts

c0dedance:
> #: Used to add a fragment identifier, which specifies a specific fragment when importing. This is often used when importing CSS stylesheets. For example: import './styles.css#section1'.

Qwik utilized this feature to load a specific export from a module.
image

c0dedance:
> This hack was drop in the last major in favor of a more clean include option. You can check this example: https://github.com/vitejs/vite-plugin-react/blob/main/playground/mdx/vite.config.ts

Thank you for your response. I don't have any further questions.

ArnaudBarre:
> Qwik utilized this feature to load a specific export from a module.

Yeah I'm not against using hash in URL for specific needs, but I think that resolutions is already complex enough and that a proper include option is a more stable way to support additional extensions

PR #375, opened 2024-10-15
fix: add `react-dom` in `optimizeOps` to handle CJS script.

### LINKED ISSUE #3114: Different imports for ReactDom
## Question

What's the idea to use two different imports for ReactDOM?
image
image

I have an issue because of that: "SyntaxError: The requested module '....node_modules/.pnpm/react-dom@18.3.1_react@18.3.1/node_modules/react-dom/index.js?v=6b12d63f' does not provide an export named 'default' (at index.mjs?v=6b12d63f:5:8)"

For usage questions, please use Stack Overflow or use [GitHub Discussions](https://github.com/radix-ui/primitives/discussions) instead of posting an issue. This helps us keep bugs and feature requests prioritized, and it increases the liklihood that more people see your question and can answer it more qiuckly!

Tiger3018:
### Description

The inclusion of `react` introduced in [vite#9056](https://github.com/vitejs/vite/pull/9056) but not for `react-dom`. Since `react-dom` is a standalone package even in React 18, we should include it without problems to have consistent behavior.

### Additional context

By merging this commit, CJS format js files like [react-dom/client.js](https://github.com/facebook/react/blob/main/packages/react-dom/npm/client.js) and [react-dom/index.js](https://github.com/facebook/react/blob/main/packages/react-dom/npm/index.js) will be enforced optimized, just like `react`'s [index.js](https://github.com/facebook/react/blob/main/packages/react/npm/index.js).

Optimization of these files will help not-well designed (import non-default exported modules `react` and `react-dom` directly) frameworks correctly import package **without developer's notice**, when using the development mode of vite, e.g. resolve [radix-ui/primitives#3114](https://github.com/radix-ui/primitives/issues/3114). More discussion of original inclusion can be seen at [vite#9052/comment](https://github.com/vitejs/vite/issues/9052#issuecomment-1182000942)

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -268,10 +268,12 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
     },
   }
 
-  // We can't add `react-dom` because the dependency is `react-dom/client`
-  // for React 18 while it's `react-dom` for React 17. We'd need to detect
-  // what React version the user has installed.
-  const dependencies = ['react', jsxImportDevRuntime, jsxImportRuntime]
+  const dependencies = [
+    'react',
+    'react-dom',
+    jsxImportDevRuntime,
+    jsxImportRuntime,
+  ]
   const staticBabelPlugins =
     typeof opts.babel === 'object' ? opts.babel?.plugins ?? [] : []
   if (hasCompilerWithDefaultRuntime(staticBabelPlugins)) {


ArnaudBarre:
@patak-dev do we have any downside except a few wasted esbuild cycles?
I don't think we have people using this plugin without React dom installed given that the Fast refresh integration relies directly on `window`

patak-cat:
Yes, I think it is a good idea to include it 👍🏼

PR #376, opened 2024-10-16
feat: using oxc transfomer

underfin:


ArnaudBarre:
I have already a plugin working in a private repo, we will keep the babel one as the default for now. 
The plugin should be plugin soon, we are waiting for the right time to announce it

PR #420, opened 2025-03-26
feat: add `reactRefreshHost` option to support module federation HMR

protoEvangelion:
### Description

This resolves HMR not working in a module federation context when using the `@module-federation/vite` plugin.

This PR simply adds a new option `reactRefreshPrefix` in a backwards compat way to allow prefixing the react-refresh url which by default is: `/@react-refresh`

The reason this works is the following example:

1. host vite server on port 3000
2. remote vite server on port 4000
3. save file in remote server
4. react-refresh triggers on remote server port 4000 but does not propagate to the host server

Allowing dev to set prefix in the plugin will force the react-refresh plugin from remote files to point at the host react-refresh module:

```
react({ reactRefreshPrefix: 'http://localhost:3000' })
```

The react refresh url becomes `http://localhost:3000/@react-refresh`.

https://github.com/user-attachments/assets/b53a5772-5002-4924-8311-28d00969f221

### Additional context

https://github.com/module-federation/vite/issues/183#issuecomment-2751825367

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/packages/common/refresh-utils.ts
+++ b/packages/common/refresh-utils.ts
@@ -21,6 +21,7 @@ export function addRefreshWrapper<M extends { mappings: string }>(
   map: M | string | typeof avoidSourceMapOption,
   pluginName: string,
   id: string,
+  reactRefreshHost = '',
 ): { code: string; map: M | null | string } {
   const hasRefresh = refreshContentRE.test(code)
   const onlyReactComp = !hasRefresh && reactCompRE.test(code)
@@ -70,7 +71,8 @@ if (import.meta.hot && !inWebWorker) {
   }
 
   const sharedHead = removeLineBreaksIfNeeded(
-    `import * as RefreshRuntime from "${runtimePublicPath}";
+    `import * as RefreshRuntime from "${reactRefreshHost}${runtimePublicPath}";
+
 const inWebWorker = typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope;
 
 `,
--- a/packages/common/refresh-utils.ts
+++ b/packages/common/refresh-utils.ts
@@ -72,7 +72,6 @@ if (import.meta.hot && !inWebWorker) {
 
   const sharedHead = removeLineBreaksIfNeeded(
     `import * as RefreshRuntime from "${reactRefreshHost}${runtimePublicPath}";
-
 const inWebWorker = typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope;
 
 `,
--- a/packages/plugin-react-swc/README.md
+++ b/packages/plugin-react-swc/README.md
@@ -93,6 +93,16 @@ react({
 })
 ```
 
+### reactRefreshHost
+
+The `reactRefreshHost` option is only necessary in a module federation context. It allows HMR to work between a remote & host server. In your remote vite config you would add your host origin:
+
+```js
+react({ reactRefreshHost: 'http://localhost:3000' })
+```
+
+Under the hood this simply updates the react refresh url from "/@react-refresh" to "http://localhost:3000/@react-refresh" to ensure you get only one Fast Refresh runtime in the whole application.
+
 ### useAtYourOwnRisk_mutateSwcOptions
 
 The future of Vite is with OXC, and from the beginning this was a design choice to not exposed too many specialties from SWC so that Vite React users can move to another transformer later.
--- a/packages/plugin-react-swc/README.md
+++ b/packages/plugin-react-swc/README.md
@@ -95,7 +95,7 @@ react({
 
 ### reactRefreshHost
 
-The `reactRefreshHost` option is only necessary in a module federation context. It enables HMR to work between a remote & host application. In your remote vite config, you would add your host origin:
+The `reactRefreshHost` option is only necessary in a module federation context. It enables HMR to work between a remote & host application. In your remote Vite config, you would add your host origin:
 
 ```js
 react({ reactRefreshHost: 'http://localhost:3000' })
--- a/packages/plugin-react-swc/README.md
+++ b/packages/plugin-react-swc/README.md
@@ -95,13 +95,13 @@ react({
 
 ### reactRefreshHost
 
-The `reactRefreshHost` option is only necessary in a module federation context. It allows HMR to work between a remote & host server. In your remote vite config you would add your host origin:
+The `reactRefreshHost` option is only necessary in a module federation context. It enables HMR to work between a remote & host application. In your remote vite config, you would add your host origin:
 
 ```js
 react({ reactRefreshHost: 'http://localhost:3000' })
 ```
 
-Under the hood this simply updates the react refresh url from "/@react-refresh" to "http://localhost:3000/@react-refresh" to ensure you get only one Fast Refresh runtime in the whole application. Note that if you define `base` you need to include that in the url provided like: `http://localhost:3000/{base}`.
+Under the hood, this simply updates the React Fash Refresh runtime URL from `/@react-refresh` to `http://localhost:3000/@react-refresh` to ensure there is only one Refresh runtime across the whole application. Note that if you define `base` option in the host application, you need to include it in the option, like: `http://localhost:3000/{base}`.
 
 ### useAtYourOwnRisk_mutateSwcOptions
 
--- a/packages/plugin-react-swc/README.md
+++ b/packages/plugin-react-swc/README.md
@@ -101,7 +101,7 @@ The `reactRefreshHost` option is only necessary in a module federation context. 
 react({ reactRefreshHost: 'http://localhost:3000' })
 ```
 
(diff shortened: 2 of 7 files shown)


protoEvangelion:
@ArnaudBarre this is my first PR here, let me know if everything looks good 😅

packages/plugin-react/src/index.ts
```diff
@@ -47,6 +47,14 @@ export interface Options {
...
     | BabelOptions
     | ((id: string, options: { ssr?: boolean }) => BabelOptions)
+  /**
+   * React refresh runtime url prefix.
+   * Useful in module federation context to enable HMR by
+   * setting the host url on a vite config which is serving a remote app.
+   * @example
+   * reactRefreshPrefix: 'http://localhost:3000'
+   */
+  reactRefreshPrefix?: string
```
ArnaudBarre: While technically you could use it to have only a path prefix, `host` makes moire sense but the intended usage 
```suggestion
  reactRefreshHost?: string
```

packages/plugin-react/src/fast-refresh.ts
```diff
@@ -1,8 +1,11 @@
 import fs from 'node:fs'
 import path from 'node:path'
 import { createRequire } from 'node:module'
+import { type Options } from './index'
 
-export const runtimePublicPath = '/@react-refresh'
+export const runtimePublicPath = (
+  reactRefreshPrefix?: Options['reactRefreshPrefix'],
+): string => `${reactRefreshPrefix ?? ''}/@react-refresh`
```
ArnaudBarre: All usages without options can keep using the `runtimePublicPath`
Idem for usages in index.ts

You can inline this template string concat in the `sharedHeader` function

packages/plugin-react/CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+Add backwards compatible option `reactRefreshPrefix` to set React refresh runtime url prefix.
+This is useful in module federation context to enable HMR by setting the host url on a vite config which is serving a remote app.
+See full discussion here: https://github.com/module-federation/vite/issues/183#issuecomment-2751825367
+
+```ts
+export default defineConfig(({ command }) => {
+  return {
+    plugins: [react({ reactRefreshPrefix: 'http://localhost:3000' })],
+  }
+})
```
ArnaudBarre: This example doesn't require the function callback, better to keep it in the simplest form
```suggestion
export default defineConfig({
  plugins: [react({ reactRefreshPrefix: 'http://localhost:3000' })],
})
```

packages/plugin-react/CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### Add backwards compatible option `reactRefreshPrefix`
+
+Add backwards compatible option `reactRefreshPrefix` to set React refresh runtime url prefix.
```
ArnaudBarre: No need to precise backwards, this is the "default" when adding feature
```suggestion
### Add `reactRefreshPrefix` option
```

packages/plugin-react/src/index.ts
```diff
@@ -47,6 +47,14 @@ export interface Options {
   babel?:
     | BabelOptions
     | ((id: string, options: { ssr?: boolean }) => BabelOptions)
+  /**
+   * React refresh runtime url prefix.
```
ArnaudBarre: Add this also to the readme

ArnaudBarre (commented):
I'm not a big fan of the complexity added by module federation, but this small feature doesn't cost much to add so why not.
@sapphi-red @patak-dev @bluwy @hi-ogawa opinions?

sapphi-red:
Adding this looks fine to me for the same reason with you. I'd like to hear from @underfin what they think about it as they have been working on MF related stuff recently.

bluwy:
Just wondering as I'm not clear of the relationship of the host & remote server, does relying on Vite's [`server.origin`](https://vite.dev/config/server-options.html#server-origin) option help here? Maybe plugin-react could generate the import that takes in account of that.

ArnaudBarre:
@bluwy Interesting, I'll try to add a test setup in the repo for this usecase so that I can try out

protoEvangelion:
Thanks for the quick feedback & good suggestions @ArnaudBarre! I just committed all your requested changes, tested, & works as expected.

@sapphi-red I'm super curious about that as well. However for those of us using module federation in a multi build system, this is still necessary. I would love to vite all the things but I'm living in a world that still webpacks all the things 😅 

@bluwy Thanks for the suggestion. I don't think that will work for assets on the remote server because if I change the origin, it will look for assets on the host server which won't be there.

protoEvangelion:
@ArnaudBarre I think we are all clear to merge this. Anything else you need me to address?

ArnaudBarre (approved):

ArnaudBarre:
@sapphi-red I let you decide when you prefer to merge given your other open PRs. I think we could also add this to other plugins

protoEvangelion:
@sapphi-red @ArnaudBarre I rebased/updated according to your last merged [pr](https://github.com/vitejs/vite-plugin-react/pull/443) moving code to common. Tested & works as expected. Side question, are you planning to do a release anytime soon? Because I haven't seen one since November 2024.

ArnaudBarre:
Thanks for updating your PR! We didn't release anything in the past months because no new bug fix or feature was merged, but we will do a release shortly after merging this.
If you want you can also add this option the SWC plugin, but this is not required to merge at all

protoEvangelion:
Makes sense 👍  I fixed conflicts again and added changes to plugin-react-swc as well.

packages/plugin-react-swc/README.md
```diff
@@ -93,6 +93,16 @@ react({
...
 
+### reactRefreshHost
+
+The `reactRefreshHost` option is only necessary in a module federation context. It allows HMR to work between a remote & host server. In your remote vite config you would add your host origin:
+
+```js
+react({ reactRefreshHost: 'http://localhost:3000' })
+```
+
+Under the hood this simply updates the react refresh url from "/@react-refresh" to "http://localhost:3000/@react-refresh" allowing HMR updates to flow from your remote to your host.
```
ArnaudBarre: THe HMR updates will still come from each dev server. The issue comes when there are multiple runtime
```suggestion
Under the hood this simply updates the react refresh url from "/@react-refresh" to "http://localhost:3000/@react-refresh" to ensure your get only one Fast Refresh runtime in the whole application.
```

packages/plugin-react-swc/src/index.ts
```diff
@@ -58,6 +58,14 @@ type Options = {
    * Exclusion of node_modules should be handled by the function if needed.
    */
   parserConfig?: (id: string) => ParserConfig | undefined
+  /**
+   * React refresh runtime url prefix.
+   * Useful in module federation context to enable HMR by
+   * setting the host url on a vite config which is serving a remote app.
```
ArnaudBarre:
```suggestion
   * setting the host url on a Vite config which is serving a remote app.
```

packages/plugin-react-swc/README.md
protoEvangelion: Add with "you" instead of "your"

protoEvangelion:
Fixed conflicts & edited readme + changelog per review comments.

ArnaudBarre (approved):

underfin:
> Adding this looks fine to me for the same reason with you. I'd like to hear from @underfin what they think about it as they have been working on MF related stuff recently.

I think it is a good idea to support mf react hmr. The host and remote could be use host react-refresh runtime. The previous bug caused by the remote react-refresh runtime is not called `RefreshRuntime.injectIntoGlobalHook(window)`. After the pr the issue is gone. Thank you @protoEvangelion nice work.

packages/common/refresh-utils.ts
```diff
@@ -70,7 +71,8 @@ if (import.meta.hot && !inWebWorker) {
   }
 
   const sharedHead = removeLineBreaksIfNeeded(
-    `import * as RefreshRuntime from "${runtimePublicPath}";
+    `import * as RefreshRuntime from "${reactRefreshHost}${runtimePublicPath}";
```
underfin, line 74: Here look like has a issue if enable vite `base` option, the vite will rewrite `/@react-refresh` to `/base/@react-refresh` at dev transformer. The host the `reactRefreshHost` is ``, so it is correct, but the remote here need to a path like `${host_url}${base}`, eg `http://localhost:3000/base`. 

I think it need to add to document.

protoEvangelion: Good callout @underfin added.

sapphi-red (approved):

yaseiva:
I would've been great if you updated the examples, I'm testing it now and every change make full reload.

Mouhanad-Dandashli:
Is there is an example of usage?

PR #423, opened 2025-03-31
refactor: merge vite plugin react swc in

sapphi-red:
Merge https://github.com/vitejs/vite-plugin-react-swc in this repo.


scripts/publishCI.ts
```diff
@@ -2,4 +2,5 @@ import { publish } from '@vitejs/release-scripts'
 
 publish({
   defaultPackage: 'plugin-react',
```
ArnaudBarre, line 4: Maybe remove default?

ArnaudBarre (approved):
Thanks !

scripts/publishCI.ts
sapphi-red: Yeah, that would be better. But the option is required, I'll make a PR to release-scripts and remove it in a separate PR.

sins921 (approved):

PR #439, opened 2025-04-03
feat: add plugin-react-oxc

sapphi-red:
### Description

A new plugin for react that leverages Oxc built into Rolldown.

#### How the test works

For each playground with `__tests__/oxc` directory, `vitestGlobalSetup.ts` copies the original playground to a separate directory for the oxc plugin (this is similar to what we have with variant playgrounds in Vite core repo, for example `playground/css/__tests__/lightningcss`). After that `vitestGlobalSetup.ts` replaces `node_modules/vite` and `node_modules/@vitejs/plugin-react` with a symlink that points to rolldown-vite and @vitejs/plugin-react-oxc.

For each playground, `vitestGlobalSetup.ts` adds `_vite-proxy.js` that re-exports `vite` (`export * from 'vite'`). This is to import vite resolved from each playground directory. In `vitestSetup.ts`, vite is imported via `_vite-proxy.js` so that rolldown-vite is used for oxc variant playgrounds.

#### Making it work with ecosystem-ci

`@vitejs/plugin-react-oxc` has `"vite": "catalog:rolldown-vite"` as a dev dependency. This catalog item points to `npm:rolldown-vite@^6.3.0-beta.5`. This is to allow ecosystem-ci to keep that dep as-is.
Because ecosystem-ci adds an override for `vite`, it also changes the `vite` devDep for `plugin-react-oxc`. To avoid that from happening, in the config for this repo for ecosystem-ci, I added `'@vitejs/plugin-react-oxc>vite': 'catalog:rolldown-vite'`. This way the devDep will keep pointing the rolldown-vite package.


sapphi-red:
I'm wondering how to make this work without overrides.

sapphi-red:
OK. I managed to make the tests working. Now I'm going to check if this works ok with ecosystem-ci.

hi-ogawa (commented):
Looks good to me!

For the testing, I was thinking about using Vitest workspace to split babel and oxc then setting up alias for each, but I realized we still need to split temp dir to avoid mixing hmr and also `@vitejs/plugin-react` needs to be aliased for vite config's esbuild to understand. Directly swapping `node_modules` seems like a reliable approach :+1: 

One question, it looks like `fast-refresh.ts` and `refreshUtils.js` are almost identical to the ones in `packages/plugin-react`. Should we try to dedupe them? I think `copyRefreshUtils.ts` script can handle just exact copy and `fast-refresh.ts` can be also replaced with a weird direct import of `../../plugin-react/src/fast-refresh`.

ArnaudBarre:
What do you think about using a setup closer to the SWC plugin for fast refresh utils?
It makes the code cleaner and doesn't change a lot (I just checked there was not anything change upstream since 01/2023)

sapphi-red:
Sure. Let me dedupe and refactor the fast refresh handling in a separate PR first.

sapphi-red:
Made a PR for https://github.com/vitejs/vite-plugin-react/pull/439#issuecomment-2781866278 here: https://github.com/vitejs/vite-plugin-react/pull/443

sapphi-red:
I've merged the main branch. The `addRefreshWrapper` is a bit complex because there's no sourcemap to modify in oxc plugin, but the overall complexity should be now reduced!

packages/plugin-react-oxc/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "type": "git",
+    "url": "git+https://github.com/vitejs/vite-plugin-react.git",
+    "directory": "packages/plugin-react-oxc"
+  },
+  "bugs": {
+    "url": "https://github.com/vitejs/vite-plugin-react/issues"
+  },
+  "homepage": "https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react#readme",
+  "dependencies": {
+    "react-refresh": "^0.14.2"
```
ArnaudBarre: Not needed anymore

packages/common/refresh-utils.ts
```diff
@@ -14,20 +14,30 @@ window.$RefreshSig$ = () => (type) => type;`
...
   const hasRefresh = refreshContentRE.test(code)
   const onlyReactComp = !hasRefresh && reactCompRE.test(code)
-  if (!hasRefresh && !onlyReactComp) return { code, map }
+  const newMap =
+    map === avoidSourceMapOption
+      ? undefined
+      : typeof map === 'string'
+        ? (JSON.parse(map) as M)
+        : map
+  if (!hasRefresh && !onlyReactComp) return { code, map: newMap }
```
ArnaudBarre: Is this wanted to return newMap in the no-op case?

packages/common/refresh-utils.ts
```diff
@@ -14,20 +14,30 @@ window.$RefreshSig$ = () => (type) => type;`
...
+  map: M | string | undefined | typeof avoidSourceMapOption,
   pluginName: string,
   id: string,
-): { code: string; map: M | string } {
+): { code: string; map: M | undefined | string } {
   const hasRefresh = refreshContentRE.test(code)
   const onlyReactComp = !hasRefresh && reactCompRE.test(code)
-  if (!hasRefresh && !onlyReactComp) return { code, map }
+  const newMap =
+    map === avoidSourceMapOption
```
ArnaudBarre: Did you try generated an empty source map that will then only be `';'.repeat(17) + newMap.mappings`?
Not a big deal but cleaner when you inspect the pipeline changes

packages/common/refresh-utils.ts
sapphi-red: Ah, my bad.

packages/common/refresh-utils.ts
sapphi-red: It's because `newMap` doesn't exist here.
https://github.com/vitejs/vite-plugin-react/pull/439/files#diff-57c3aa7ebcc7d479870d1da730fee916a15a41215f2a6af821383235f62f3b59R114-R129
We can generate it by `new MagicString(code).generateMap({ hires: 'boundary' })`, but I guess that has a bigger overhead.

ArnaudBarre: Does rollup allows to get the current sourceMap of the transform pipeline?
So we can edit it instead of generating a new one and having Vite merge them after

sapphi-red: It has a way to get the current combined source map ([`this.getCombinedSourcemap`](https://rollupjs.org/plugin-development/#this-getcombinedsourcemap)). But there isn't a way to tell rollup that the returned sourcemap is a combined one.

ArnaudBarre: I did some experiment but using an empty sourceMappings where only `;` is apprend breaks the concatenation somewhere because the output doesn't contain any source map (without logs)

So I tried this:

```ts
        const { code: newCode, map } = addRefreshWrapper(
          code,
          {
            version: 3,
            file: undefined,
            sources: [''],
            sourcesContent: undefined,
            names: [],
            mappings: 'AAAA;'.repeat(code.split('\n').length - 1),
          },
          '@vitejs/plugin-react-oxc',
          id,
        )
        return { code: newCode, map }
```

Inspired by what magic-string does as a default and once again was surprised out slow unpredictable [JS perf is](https://perf.link/#eyJpZCI6Ingyems3ZTA0OGR3IiwidGl0bGUiOiJDb3VudCBsaW5lIGJyZWFrIGluIGEgc3RyaW5nIiwiYmVmb3JlIjoiY29uc3QgZGF0YSA9IGBjb25zb2xlLmxvZygnQXBwLmpzeCAxJykgLy8gZm9yIHNvdXJjZW1hcFxuZnVuY3Rpb24gQXBwKCkge1xuICByZXR1cm4gKFxuICAgIDxkaXZcbiAgICAgIG9uQ2xpY2s9eygpID0%2BIHtcbiAgICAgICAgdGhyb3cgbmV3IEVycm9yKCdiYXInKVxuICAgICAgfX1cbiAgICA%2BXG4gICAgICBmb29cbiAgICA8L2Rpdj5cbiAgKVxufVxuXG5jb25zb2xlLmxvZygnQXBwLmpzeCAyJykgLy8gZm9yIHNvdXJjZW1hcFxuXG5leHBvcnQgZGVmYXVsdCBBcHBcbmBcbmNvbnN0IGN1c3RvbUNvdW50ID0gKHN0cikgPT4ge1xuIGxldCBjb3VudCA9IDA7XG4gZm9yIChjb25zdCBjaGFyIG9mIHN0cikgaWYgKGNoYXIgPT09IFwiXFxuXCIpIGNvdW50KytcbiByZXR1cm4gY291bnRcbn0iLCJ0ZXN0cyI6W3sibmFtZSI6InNwbGl0IiwiY29kZSI6ImRhdGEuc3BsaXQoJ1xcbicpLmxlbmd0aCAtIDE7IiwicnVucyI6Wzg4NjcwMDAsMTkxMzAwMCwyNDE4MDAwLDQ3MTAwMDAsNTAxOTAwMCw0Njc1MDAwLDIzMzUwMDAsMTM4OTAwMCwyNjg0MDAwLDEwNzkwMDAwLDMxMTgwMDAsNTEyOTAwMCwyNjkwMDAwLDExODgwMDAsMTE4ODAwMCwxMDgzODAwMCwxMTg4MDAwLDExODgwMDAsOTMzNDAwMCwxMTg4MDAwLDI2MjgwMDAsNjc2OTAwMCwzODkxMDAwLDExODgwMDAsNDUyMDAwMCwxMTg4MDAwLDUyMzAwMDAsOTk0MDAwMCwxMTY0MDAwLDEwODY1MDAwLDgxMDAwLDEyNzI0MDAwLDE4MTQwMDAsMTA0MjYwMDAsMTk0NjAwMCw5NTMyMDAwLDEyNjAzMDAwLDExODgwMDAsMTYxNzAwMCwxMTMwNjAwMCwxNDAwMDAwLDQzOTYwMDAsNTEyMDAwLDc5NjAwMCw0MjQ0MDAwLDEwODg0MDAwLDczOTgwMDAsMjQxNzAwMCwzMDEyMDAwLDEyMTAyMDAwLDIxNTcwMDAsOTUwNDAwMCw0NzI1MDAwLDQyNjMwMDAsMjQ1ODAwMCw3MTYwMDAsNDM3OTAwMCwxMDUyMTAwMCw1NDU1MDAwLDM1NjEwMDAsNTY3MTAwMCwxMTg4MDAwLDExODgwMDAsMTA2MTIwMDAsNDA3ODAwMCwyNTQyMDAwLDU2MDQwMDAsMTE4ODAwMCwzMDY1MDAwLDgxNDkwMDAsOTYwMzAwMCw0MzM1MDAwLDQzODcwMDAsMTE4ODAwMCwzNDEwMDAwLDc1MzkwMDAsMTAyMTcwMDAsNDY5NTAwMCwxMTg4MDAwLDEyMTk1MDAwLDMxODkwMDAsODI3NTAwMCwxMTQ2MDAwMCwzMDk5MDAwLDI4MzkwMDAsMTEzMzgwMDAsNTAzMTAwMCw3ODc3MDAwLDIzNzgwMDAsNDE2MzAwMCwxMTIxMTAwMCw0NTc1MDAwLDExODgwMDAsMjg4MDAwLDk3MTYwMDAsMTI0MDEwMDAsOTAxMDAwMCwzMDU5MDAwLDU2NzUwMDAsMjYwMDAwMF0sIm9wcyI6NTA4NzgzMH0seyJuYW1lIjoiUmVnZXgiLCJjb2RlIjoiZGF0YS5tYXRjaCgvXFxuL2cpPy5sZW5ndGggPz8gMCIsInJ1bnMiOls0MzA0MDAwLDExMTAwMDAsMTQ1MDAwMCwyMjI5MDAwLDE3MTcwMDAsMjI0NjAwMCwxNDY2MDAwLDg4MTAwMCwxNTI4MDAwLDQ2NzcwMDAsMTY2NzAwMCwyNDE1MDAwLDE1NzEwMDAsNjc5MDAwLDczMzAwMCw0Njc0MDAwLDUyMzAwMCwyODEwMDAsMjgxMjAwMCw1NjYwMDAsMTE0ODAwMCwzMzU4MDAwLDE5OTgwMDAsODA2MDAwLDEzMzYwMDAsOTIzMDAwLDE3MzgwMDAsNDEwNzAwMCw0MzAwMCw0NjQxMDAwLDUzMzkwMDAsMTEwMzAwMCw1MzMwMDAsMzAyMDAwMCwxMTAzMDAwLDMzOTMwMDAsNTIyNzAwMCwyODEwMDAsMzgyMDAwLDUxMTcwMDAsOTYyMDAwLDEzMjkwMDAsMjU5MDAwLDI4NTAwMCwxNDE5MDAwLDQ2MjUwMDAsMzE3OTAwMCw2NjcwMDAsMTQ5NzAwMCw1Njg0MDAwLDExMTgwMDAsMjk3MDAwMCwxNzE3MDAwLDE0NjYwMDAsMTEwMzAwMCw2OTYwMDAsMTM1ODAwMCwzODkyMDAwLDIyMTcwMDAsMTgzMTAwMCwyNTAyMDAwLDI5NjAwMCwxMTAzMDAwLDQ1MjEwMDAsMjEwMjAwMCwxMzAyMDAwLDE4NjgwMDAsNjkzMDAwLDExMDMwMDAsMjYxNTAwMCwzMDM3MDAwLDE3ODgwMDAsMjA2NzAwMCw3NTEwMDAsMTY1MjAwMCwzMjU0MDAwLDQzMjEwMDAsMTkwNjAwMCw4NTgwMDAsMzM3MDAwLDE2MjgwMDAsMzAxOTAwMCw0MzExMDAwLDEyMDUwMDAsMTQ4MTAwMCwzOTE0MDAwLDE0NDQwMDAsMjI0NDAwMCwxMTQ3MDAwLDE5MDIwMDAsNTA0ODAwMCwyMjM3MDAwLDQxMDAwMCwyMjAwMCw0NTYyMDAwLDUwNzcwMDAsMjIyODAwMCwxNDUxMDAwLDE2MjQwMDAsMTEwMzAwMF0sIm9wcyI6MjA1NTMyMH0seyJuYW1lIjoiSXRlcmF0ZSBvbiBzdHJpbmciLCJjb2RlIjoiY3VzdG9tQ291bnQoZGF0YSkiLCJydW5zIjpbMTIwMDAwLDIxMDAwLDgwMDAsMjQwMDAsNDcwMDAsNDEwMDAsNDAwMCwxNDAwMCwyNTAwMCwyMDUwMDAsMzUwMDAsNzMwMDAsMzMwMDAsMTAwMDAsMTgxMDAwLDE5MTAwMCw4MDAwLDEwMDAsMTI1MDAwLDkwMDAsNzAwMCwxMjgwMDAsMjMwMDAsMTIwMDAsMzEwMDAsMTYwMDAsNTIwMDAsMTM4MDAwLDIwMDAsNDIwMDAsMTkyMDAwLDEzMDAwLDgwMDAsOTgwMDAsODAwMCwxNTYwMDAsMTcyMDAwLDEwMDAwLDgwMDAsMjA3MDAwLDE1MDAwLDM2MDAwLDMwMDAsNzAwMCwyNTAwMCwxNDUwMDAsODkwMDAsNDAwMCwxMjAwMCwyMDcwMDAsMTkwMDAsMTE2MDAwLDU4MDAwLDQ3MDAwLDExMDAwLDExMDAwLDMwMDAwLDE5MDAwMCw0MTAwMCw1MjAwMCw0NzAwMCw4MDAwLDE5ODAwMCwxNTYwMDAsMjUwMDAsODAwMCwyMjAwMCw4MDAwLDgwMDAsOTAwMDAsMTI4MDAwLDQ2MDAwLDIyMDAwLDgwMDAsMjMwMDAsMTE1MDAwLDE3NzAwMCw2MjAwMCw0MDAwLDEzMDAwLDE5MDAwLDEyMDAwMCwxNDgwMDAsMzYwMDAsMTMwMDAsMTk3MDAwLDUxMDAwLDExNjAwMCwyNTAwMCw1MTAwMCwxMDUwMDAsNzAwMDAsODAwMCwzMDAwLDEyMDAwMCwxNzMwMDAsMTA0MDAwLDgwMDAsNDcwMDAsMjIwMDBdLCJvcHMiOjYyMjAwfV0sInVwZGF0ZWQiOiIyMDI1LTA0LTA3VDE5OjMxOjA5LjY3MloifQ%3D%3D).

This test also made me find that the `';'.repeat` should be 16 and not 17 because there are 17 lines but 16 line breaks.

With the change I have this kind of [output](https://evanw.github.io/source-map-visualization/#Mjg1NQBpbXBvcnQgeyBjcmVhdGVIb3RDb250ZXh0IGFzIF9fdml0ZV9fY3JlYXRlSG90Q29udGV4dCB9IGZyb20gIi9Adml0ZS9jbGllbnQiO2ltcG9ydC5tZXRhLmhvdCA9IF9fdml0ZV9fY3JlYXRlSG90Q29udGV4dCgiL0FwcC5qc3giKTtpbXBvcnQgKiBhcyBSZWZyZXNoUnVudGltZSBmcm9tICIvQHJlYWN0LXJlZnJlc2giOwpjb25zdCBpbldlYldvcmtlciA9IHR5cGVvZiBXb3JrZXJHbG9iYWxTY29wZSAhPT0gJ3VuZGVmaW5lZCcgJiYgc2VsZiBpbnN0YW5jZW9mIFdvcmtlckdsb2JhbFNjb3BlOwoKbGV0IHByZXZSZWZyZXNoUmVnOwpsZXQgcHJldlJlZnJlc2hTaWc7CgppZiAoaW1wb3J0Lm1ldGEuaG90ICYmICFpbldlYldvcmtlcikgewogIGlmICghd2luZG93LiRSZWZyZXNoUmVnJCkgewogICAgdGhyb3cgbmV3IEVycm9yKAogICAgICAiQHZpdGVqcy9wbHVnaW4tcmVhY3Qtb3hjIGNhbid0IGRldGVjdCBwcmVhbWJsZS4gU29tZXRoaW5nIGlzIHdyb25nLiIKICAgICk7CiAgfQoKICBwcmV2UmVmcmVzaFJlZyA9IHdpbmRvdy4kUmVmcmVzaFJlZyQ7CiAgcHJldlJlZnJlc2hTaWcgPSB3aW5kb3cuJFJlZnJlc2hTaWckOwogIHdpbmRvdy4kUmVmcmVzaFJlZyQgPSBSZWZyZXNoUnVudGltZS5nZXRSZWZyZXNoUmVnKCIvVXNlcnMvYXJuYXVkL2dpdC92aXRlLXBsdWdpbi1yZWFjdC9wbGF5Z3JvdW5kL3JlYWN0LXNvdXJjZW1hcC9BcHAuanN4Iik7CiAgd2luZG93LiRSZWZyZXNoU2lnJCA9IFJlZnJlc2hSdW50aW1lLmNyZWF0ZVNpZ25hdHVyZUZ1bmN0aW9uRm9yVHJhbnNmb3JtOwp9Cgp2YXIgX2pzeEZpbGVOYW1lID0gIi9Vc2Vycy9hcm5hdWQvZ2l0L3ZpdGUtcGx1Z2luLXJlYWN0L3BsYXlncm91bmQvcmVhY3Qtc291cmNlbWFwL0FwcC5qc3giOwppbXBvcnQgX192aXRlX19janNJbXBvcnQyX3JlYWN0X2pzeERldlJ1bnRpbWUgZnJvbSAiL25vZGVfbW9kdWxlcy8udml0ZS9kZXBzL3JlYWN0X2pzeC1kZXYtcnVudGltZS5qcz92PTlhN2Y2YjBkIjsgY29uc3QgX2pzeERFViA9IF9fdml0ZV9fY2pzSW1wb3J0Ml9yZWFjdF9qc3hEZXZSdW50aW1lWyJqc3hERVYiXTsKY29uc29sZS5sb2coIkFwcC5qc3ggMSIpOwpmdW5jdGlvbiBBcHAoKSB7CglyZXR1cm4gLyogQF9fUFVSRV9fICovIF9qc3hERVYoImRpdiIsIHsKCQlvbkNsaWNrOiAoKSA9PiB7CgkJCXRocm93IG5ldyBFcnJvcigiYmFyIik7CgkJfSwKCQljaGlsZHJlbjogImZvbyIKCX0sIHZvaWQgMCwgZmFsc2UsIHsKCQlmaWxlTmFtZTogX2pzeEZpbGVOYW1lLAoJCWxpbmVOdW1iZXI6IDQsCgkJY29sdW1uTnVtYmVyOiA1Cgl9LCB0aGlzKTsKfQpfYyA9IEFwcDsKY29uc29sZS5sb2coIkFwcC5qc3ggMiIpOwpleHBvcnQgZGVmYXVsdCBBcHA7CnZhciBfYzsKJFJlZnJlc2hSZWckKF9jLCAiQXBwIik7CgoKaWYgKGltcG9ydC5tZXRhLmhvdCAmJiAhaW5XZWJXb3JrZXIpIHsKICB3aW5kb3cuJFJlZnJlc2hSZWckID0gcHJldlJlZnJlc2hSZWc7CiAgd2luZG93LiRSZWZyZXNoU2lnJCA9IHByZXZSZWZyZXNoU2lnOwp9CgoKaWYgKGltcG9ydC5tZXRhLmhvdCAmJiAhaW5XZWJXb3JrZXIpIHsKICBSZWZyZXNoUnVudGltZS5fX2htcl9pbXBvcnQoaW1wb3J0Lm1ldGEudXJsKS50aGVuKChjdXJyZW50RXhwb3J0cykgPT4gewogICAgUmVmcmVzaFJ1bnRpbWUucmVnaXN0ZXJFeHBvcnRzRm9yUmVhY3RSZWZyZXNoKCIvVXNlcnMvYXJuYXVkL2dpdC92aXRlLXBsdWdpbi1yZWFjdC9wbGF5Z3JvdW5kL3JlYWN0LXNvdXJjZW1hcC9BcHAuanN4IiwgY3VycmVudEV4cG9ydHMpOwogICAgaW1wb3J0Lm1ldGEuaG90LmFjY2VwdCgobmV4dEV4cG9ydHMpID0+IHsKICAgICAgaWYgKCFuZXh0RXhwb3J0cykgcmV0dXJuOwogICAgICBjb25zdCBpbnZhbGlkYXRlTWVzc2FnZSA9IFJlZnJlc2hSdW50aW1lLnZhbGlkYXRlUmVmcmVzaEJvdW5kYXJ5QW5kRW5xdWV1ZVVwZGF0ZSgiL1VzZXJzL2FybmF1ZC9naXQvdml0ZS1wbHVnaW4tcmVhY3QvcGxheWdyb3VuZC9yZWFjdC1zb3VyY2VtYXAvQXBwLmpzeCIsIGN1cnJlbnRFeHBvcnRzLCBuZXh0RXhwb3J0cyk7CiAgICAgIGlmIChpbnZhbGlkYXRlTWVzc2FnZSkgaW1wb3J0Lm1ldGEuaG90LmludmFsaWRhdGUoaW52YWxpZGF0ZU1lc3NhZ2UpOwogICAgfSk7CiAgfSk7Cn0KCi8vIyBzb3VyY2VNYXBwaW5nVVJMPWRhdGE6YXBwbGljYXRpb24vanNvbjtiYXNlNjQsZXlKMlpYSnphVzl1SWpvekxDSnRZWEJ3YVc1bmN5STZJanM3T3pzN096czdPenM3T3pzN096czdPenM3TzBGQlFVRTdRVUZEUVR0QlFVTkZPenRCUVVkTk96czdPenM3T3p0QlFVMVFPenRCUVVWRU8wRkJSVUVpTENKdVlXMWxjeUk2VzEwc0ltbG5ibTl5WlV4cGMzUWlPbHRkTENKemIzVnlZMlZ6SWpwYklrRndjQzVxYzNnaVhTd2ljMjkxY21ObGMwTnZiblJsYm5RaU9sc2lZMjl1YzI5c1pTNXNiMmNvSjBGd2NDNXFjM2dnTVNjcElDOHZJR1p2Y2lCemIzVnlZMlZ0WVhCY2JtWjFibU4wYVc5dUlFRndjQ2dwSUh0Y2JpQWdjbVYwZFhKdUlDaGNiaUFnSUNBOFpHbDJYRzRnSUNBZ0lDQnZia05zYVdOclBYc29LU0E5UGlCN1hHNGdJQ0FnSUNBZ0lIUm9jbTkzSUc1bGR5QkZjbkp2Y2lnblltRnlKeWxjYmlBZ0lDQWdJSDE5WEc0Z0lDQWdQbHh1SUNBZ0lDQWdabTl2WEc0Z0lDQWdQQzlrYVhZK1hHNGdJQ2xjYm4xY2JseHVZMjl1YzI5c1pTNXNiMmNvSjBGd2NDNXFjM2dnTWljcElDOHZJR1p2Y2lCemIzVnlZMlZ0WVhCY2JseHVaWGh3YjNKMElHUmxabUYxYkhRZ1FYQndYRzRpWFN3aVptbHNaU0k2SWk5VmMyVnljeTloY201aGRXUXZaMmwwTDNacGRHVXRjR3gxWjJsdUxYSmxZV04wTDNCc1lYbG5jbTkxYm1RdmNtVmhZM1F0YzI5MWNtTmxiV0Z3TDBGd2NDNXFjM2dpZlE9PTQ5NgB7InZlcnNpb24iOjMsIm1hcHBpbmdzIjoiOzs7Ozs7Ozs7Ozs7Ozs7Ozs7Ozs7QUFBQTtBQUNBO0FBQ0U7O0FBR007Ozs7Ozs7O0FBTVA7O0FBRUQ7QUFFQSIsIm5hbWVzIjpbXSwiaWdub3JlTGlzdCI6W10sInNvdXJjZXMiOlsiQXBwLmpzeCJdLCJzb3VyY2VzQ29udGVudCI6WyJjb25zb2xlLmxvZygnQXBwLmpzeCAxJykgLy8gZm9yIHNvdXJjZW1hcFxuZnVuY3Rpb24gQXBwKCkge1xuICByZXR1cm4gKFxuICAgIDxkaXZcbiAgICAgIG9uQ2xpY2s9eygpID0+IHtcbiAgICAgICAgdGhyb3cgbmV3IEVycm9yKCdiYXInKVxuICAgICAgfX1cbiAgICA+XG4gICAgICBmb29cbiAgICA8L2Rpdj5cbiAgKVxufVxuXG5jb25zb2xlLmxvZygnQXBwLmpzeCAyJykgLy8gZm9yIHNvdXJjZW1hcFxuXG5leHBvcnQgZGVmYXVsdCBBcHBcbiJdLCJmaWxlIjoiL1VzZXJzL2FybmF1ZC9naXQvdml0ZS1wbHVnaW4tcmVhY3QvcGxheWdyb3VuZC9yZWFjdC1zb3VyY2VtYXAvQXBwLmpzeCJ9).

For [Babel](https://evanw.github.io/source-map-visualization/#MzQzMwBpbXBvcnQgeyBjcmVhdGVIb3RDb250ZXh0IGFzIF9fdml0ZV9fY3JlYXRlSG90Q29udGV4dCB9IGZyb20gIi9Adml0ZS9jbGllbnQiO2ltcG9ydC5tZXRhLmhvdCA9IF9fdml0ZV9fY3JlYXRlSG90Q29udGV4dCgiL0FwcC5qc3giKTtpbXBvcnQgX192aXRlX19janNJbXBvcnQwX3JlYWN0X2pzeERldlJ1bnRpbWUgZnJvbSAiL25vZGVfbW9kdWxlcy8udml0ZS9kZXBzL3JlYWN0X2pzeC1kZXYtcnVudGltZS5qcz92PTk2MTVlNjNlIjsgY29uc3QganN4REVWID0gX192aXRlX19janNJbXBvcnQwX3JlYWN0X2pzeERldlJ1bnRpbWVbImpzeERFViJdOwppbXBvcnQgUmVmcmVzaFJ1bnRpbWUgZnJvbSAiL0ByZWFjdC1yZWZyZXNoIjsKY29uc3QgaW5XZWJXb3JrZXIgPSB0eXBlb2YgV29ya2VyR2xvYmFsU2NvcGUgIT09ICJ1bmRlZmluZWQiICYmIHNlbGYgaW5zdGFuY2VvZiBXb3JrZXJHbG9iYWxTY29wZTsKbGV0IHByZXZSZWZyZXNoUmVnOwpsZXQgcHJldlJlZnJlc2hTaWc7CmlmIChpbXBvcnQubWV0YS5ob3QgJiYgIWluV2ViV29ya2VyKSB7CiAgaWYgKCF3aW5kb3cuX192aXRlX3BsdWdpbl9yZWFjdF9wcmVhbWJsZV9pbnN0YWxsZWRfXykgewogICAgdGhyb3cgbmV3IEVycm9yKCJAdml0ZWpzL3BsdWdpbi1yZWFjdCBjYW4ndCBkZXRlY3QgcHJlYW1ibGUuIFNvbWV0aGluZyBpcyB3cm9uZy4gU2VlIGh0dHBzOi8vZ2l0aHViLmNvbS92aXRlanMvdml0ZS1wbHVnaW4tcmVhY3QvcHVsbC8xMSNkaXNjdXNzaW9uX3I0MzA4NzkyMDEiKTsKICB9CiAgcHJldlJlZnJlc2hSZWcgPSB3aW5kb3cuJFJlZnJlc2hSZWckOwogIHByZXZSZWZyZXNoU2lnID0gd2luZG93LiRSZWZyZXNoU2lnJDsKICB3aW5kb3cuJFJlZnJlc2hSZWckID0gKHR5cGUsIGlkKSA9PiB7CiAgICBSZWZyZXNoUnVudGltZS5yZWdpc3Rlcih0eXBlLCAiL1VzZXJzL2FybmF1ZC9naXQvdml0ZS1wbHVnaW4tcmVhY3QvcGxheWdyb3VuZC9yZWFjdC1zb3VyY2VtYXAvQXBwLmpzeCAiICsgaWQpOwogIH07CiAgd2luZG93LiRSZWZyZXNoU2lnJCA9IFJlZnJlc2hSdW50aW1lLmNyZWF0ZVNpZ25hdHVyZUZ1bmN0aW9uRm9yVHJhbnNmb3JtOwp9CmNvbnNvbGUubG9nKCJBcHAuanN4IDEiKTsKZnVuY3Rpb24gQXBwKCkgewogIHJldHVybiAvKiBAX19QVVJFX18gKi8ganN4REVWKAogICAgImRpdiIsCiAgICB7CiAgICAgIG9uQ2xpY2s6ICgpID0+IHsKICAgICAgICB0aHJvdyBuZXcgRXJyb3IoImJhciIpOwogICAgICB9LAogICAgICBjaGlsZHJlbjogImZvbyIKICAgIH0sCiAgICB2b2lkIDAsCiAgICBmYWxzZSwKICAgIHsKICAgICAgZmlsZU5hbWU6ICIvVXNlcnMvYXJuYXVkL2dpdC92aXRlLXBsdWdpbi1yZWFjdC9wbGF5Z3JvdW5kL3JlYWN0LXNvdXJjZW1hcC9BcHAuanN4IiwKICAgICAgbGluZU51bWJlcjogNCwKICAgICAgY29sdW1uTnVtYmVyOiA1CiAgICB9LAogICAgdGhpcwogICk7Cn0KX2MgPSBBcHA7CmNvbnNvbGUubG9nKCJBcHAuanN4IDIiKTsKZXhwb3J0IGRlZmF1bHQgQXBwOwp2YXIgX2M7CiRSZWZyZXNoUmVnJChfYywgIkFwcCIpOwppZiAoaW1wb3J0Lm1ldGEuaG90ICYmICFpbldlYldvcmtlcikgewogIHdpbmRvdy4kUmVmcmVzaFJlZyQgPSBwcmV2UmVmcmVzaFJlZzsKICB3aW5kb3cuJFJlZnJlc2hTaWckID0gcHJldlJlZnJlc2hTaWc7Cn0KaWYgKGltcG9ydC5tZXRhLmhvdCAmJiAhaW5XZWJXb3JrZXIpIHsKICBSZWZyZXNoUnVudGltZS5fX2htcl9pbXBvcnQoaW1wb3J0Lm1ldGEudXJsKS50aGVuKChjdXJyZW50RXhwb3J0cykgPT4gewogICAgUmVmcmVzaFJ1bnRpbWUucmVnaXN0ZXJFeHBvcnRzRm9yUmVhY3RSZWZyZXNoKCIvVXNlcnMvYXJuYXVkL2dpdC92aXRlLXBsdWdpbi1yZWFjdC9wbGF5Z3JvdW5kL3JlYWN0LXNvdXJjZW1hcC9BcHAuanN4IiwgY3VycmVudEV4cG9ydHMpOwogICAgaW1wb3J0Lm1ldGEuaG90LmFjY2VwdCgobmV4dEV4cG9ydHMpID0+IHsKICAgICAgaWYgKCFuZXh0RXhwb3J0cykgcmV0dXJuOwogICAgICBjb25zdCBpbnZhbGlkYXRlTWVzc2FnZSA9IFJlZnJlc2hSdW50aW1lLnZhbGlkYXRlUmVmcmVzaEJvdW5kYXJ5QW5kRW5xdWV1ZVVwZGF0ZSgiL1VzZXJzL2FybmF1ZC9naXQvdml0ZS1wbHVnaW4tcmVhY3QvcGxheWdyb3VuZC9yZWFjdC1zb3VyY2VtYXAvQXBwLmpzeCIsIGN1cnJlbnRFeHBvcnRzLCBuZXh0RXhwb3J0cyk7CiAgICAgIGlmIChpbnZhbGlkYXRlTWVzc2FnZSkgaW1wb3J0Lm1ldGEuaG90LmludmFsaWRhdGUoaW52YWxpZGF0ZU1lc3NhZ2UpOwogICAgfSk7CiAgfSk7Cn0KCi8vIyBzb3VyY2VNYXBwaW5nVVJMPWRhdGE6YXBwbGljYXRpb24vanNvbjtiYXNlNjQsZXlKMlpYSnphVzl1SWpvekxDSnRZWEJ3YVc1bmN5STZJa0ZCUjBrN1FVRklTa0VzVDBGQlR5eHZRa0ZCYTBJN1FVRkJRVHRCUVVGQk8wRkJRVUU3UVVGQlFUdEJRVUZCTzBGQlFVRTdRVUZCUVR0QlFVRkJPMEZCUVVFN1FVRkJRVHRCUVVGQk8wRkJRVUU3UVVGQlFUdEJRVUZCTzBGQlFVRTdRVUZEZWtJc1UwRkJVME1zVFVGQlRUdEJRVU5pTEZOQlEwVTdRVUZCUVN4SlFVRkRPMEZCUVVFN1FVRkJRU3hOUVVORExGTkJRVk1zVFVGQlRUdEJRVU5pTEdOQlFVMHNTVUZCU1VNc1RVRkJUU3hMUVVGTE8wRkJRVUVzVFVGRGRrSTdRVUZCUVN4TlFVRkZPMEZCUVVFN1FVRkJRU3hKUVVoS08wRkJRVUU3UVVGQlFUdEJRVUZCTzBGQlFVRTdRVUZCUVR0QlFVRkJPMEZCUVVFN1FVRkJRU3hGUVUxQk8wRkJSVW83UVVGQlEwTXNTMEZXVVVZN1FVRlpWRVFzVVVGQlVVa3NTVUZCU1N4WFFVRlhPMEZCUlhaQ0xHVkJRV1ZJTzBGQlFVY3NTVUZCUVVVN1FVRkJRVVVzWVVGQlFVWXNTVUZCUVNJc0ltNWhiV1Z6SWpwYkltTnZibk52YkdVaUxDSkJjSEFpTENKRmNuSnZjaUlzSWw5aklpd2liRzluSWl3aUpGSmxabkpsYzJoU1pXY2tJbDBzSW1sbmJtOXlaVXhwYzNRaU9sdGRMQ0p6YjNWeVkyVnpJanBiSWtGd2NDNXFjM2dpWFN3aWMyOTFjbU5sYzBOdmJuUmxiblFpT2xzaVkyOXVjMjlzWlM1c2IyY29KMEZ3Y0M1cWMzZ2dNU2NwSUM4dklHWnZjaUJ6YjNWeVkyVnRZWEJjYm1aMWJtTjBhVzl1SUVGd2NDZ3BJSHRjYmlBZ2NtVjBkWEp1SUNoY2JpQWdJQ0E4WkdsMlhHNGdJQ0FnSUNCdmJrTnNhV05yUFhzb0tTQTlQaUI3WEc0Z0lDQWdJQ0FnSUhSb2NtOTNJRzVsZHlCRmNuSnZjaWduWW1GeUp5bGNiaUFnSUNBZ0lIMTlYRzRnSUNBZ1BseHVJQ0FnSUNBZ1ptOXZYRzRnSUNBZ1BDOWthWFkrWEc0Z0lDbGNibjFjYmx4dVkyOXVjMjlzWlM1c2IyY29KMEZ3Y0M1cWMzZ2dNaWNwSUM4dklHWnZjaUJ6YjNWeVkyVnRZWEJjYmx4dVpYaHdiM0owSUdSbFptRjFiSFFnUVhCd1hHNGlYU3dpWm1sc1pTSTZJaTlWYzJWeWN5OWhjbTVoZFdRdloybDBMM1pwZEdVdGNHeDFaMmx1TFhKbFlXTjBMM0JzWVhsbmNtOTFibVF2Y21WaFkzUXRjMjkxY21ObGJXRndMMEZ3Y0M1cWMzZ2lmUT09ODI2AHsidmVyc2lvbiI6MywibWFwcGluZ3MiOiJBQUdJO0FBSEpBLE9BQU8sb0JBQWtCO0FBQUE7QUFBQTtBQUFBO0FBQUE7QUFBQTtBQUFBO0FBQUE7QUFBQTtBQUFBO0FBQUE7QUFBQTtBQUFBO0FBQUE7QUFBQTtBQUFBO0FBQ3pCLFNBQVNDLE1BQU07QUFDYixTQUNFO0FBQUEsSUFBQztBQUFBO0FBQUEsTUFDQyxTQUFTLE1BQU07QUFDYixjQUFNLElBQUlDLE1BQU0sS0FBSztBQUFBLE1BQ3ZCO0FBQUEsTUFBRTtBQUFBO0FBQUEsSUFISjtBQUFBO0FBQUE7QUFBQTtBQUFBO0FBQUE7QUFBQTtBQUFBO0FBQUEsRUFNQTtBQUVKO0FBQUNDLEtBVlFGO0FBWVRELFFBQVFJLElBQUksV0FBVztBQUV2QixlQUFlSDtBQUFHLElBQUFFO0FBQUFFLGFBQUFGLElBQUEiLCJuYW1lcyI6WyJjb25zb2xlIiwiQXBwIiwiRXJyb3IiLCJfYyIsImxvZyIsIiRSZWZyZXNoUmVnJCJdLCJpZ25vcmVMaXN0IjpbXSwic291cmNlcyI6WyJBcHAuanN4Il0sInNvdXJjZXNDb250ZW50IjpbImNvbnNvbGUubG9nKCdBcHAuanN4IDEnKSAvLyBmb3Igc291cmNlbWFwXG5mdW5jdGlvbiBBcHAoKSB7XG4gIHJldHVybiAoXG4gICAgPGRpdlxuICAgICAgb25DbGljaz17KCkgPT4ge1xuICAgICAgICB0aHJvdyBuZXcgRXJyb3IoJ2JhcicpXG4gICAgICB9fVxuICAgID5cbiAgICAgIGZvb1xuICAgIDwvZGl2PlxuICApXG59XG5cbmNvbnNvbGUubG9nKCdBcHAuanN4IDInKSAvLyBmb3Igc291cmNlbWFwXG5cbmV4cG9ydCBkZWZhdWx0IEFwcFxuIl0sImZpbGUiOiIvVXNlcnMvYXJuYXVkL2dpdC92aXRlLXBsdWdpbi1yZWFjdC9wbGF5Z3JvdW5kL3JlYWN0LXNvdXJjZW1hcC9BcHAuanN4In0=) I found that's it's a bit broken.

For [SWC](https://evanw.github.io/source-map-visualization/#MzAxMgBpbXBvcnQgeyBjcmVhdGVIb3RDb250ZXh0IGFzIF9fdml0ZV9fY3JlYXRlSG90Q29udGV4dCB9IGZyb20gIi9Adml0ZS9jbGllbnQiO2ltcG9ydC5tZXRhLmhvdCA9IF9fdml0ZV9fY3JlYXRlSG90Q29udGV4dCgiL0FwcC5qc3giKTtpbXBvcnQgKiBhcyBSZWZyZXNoUnVudGltZSBmcm9tICIvQHJlYWN0LXJlZnJlc2giOwpjb25zdCBpbldlYldvcmtlciA9IHR5cGVvZiBXb3JrZXJHbG9iYWxTY29wZSAhPT0gJ3VuZGVmaW5lZCcgJiYgc2VsZiBpbnN0YW5jZW9mIFdvcmtlckdsb2JhbFNjb3BlOwoKbGV0IHByZXZSZWZyZXNoUmVnOwpsZXQgcHJldlJlZnJlc2hTaWc7CgppZiAoaW1wb3J0Lm1ldGEuaG90ICYmICFpbldlYldvcmtlcikgewogIGlmICghd2luZG93LiRSZWZyZXNoUmVnJCkgewogICAgdGhyb3cgbmV3IEVycm9yKAogICAgICAiQHZpdGVqcy9wbHVnaW4tcmVhY3Qtc3djIGNhbid0IGRldGVjdCBwcmVhbWJsZS4gU29tZXRoaW5nIGlzIHdyb25nLiIKICAgICk7CiAgfQoKICBwcmV2UmVmcmVzaFJlZyA9IHdpbmRvdy4kUmVmcmVzaFJlZyQ7CiAgcHJldlJlZnJlc2hTaWcgPSB3aW5kb3cuJFJlZnJlc2hTaWckOwogIHdpbmRvdy4kUmVmcmVzaFJlZyQgPSBSZWZyZXNoUnVudGltZS5nZXRSZWZyZXNoUmVnKCIvVXNlcnMvYXJuYXVkL2dpdC92aXRlLXBsdWdpbi1yZWFjdC9wbGF5Z3JvdW5kL3JlYWN0LXNvdXJjZW1hcC9BcHAuanN4Iik7CiAgd2luZG93LiRSZWZyZXNoU2lnJCA9IFJlZnJlc2hSdW50aW1lLmNyZWF0ZVNpZ25hdHVyZUZ1bmN0aW9uRm9yVHJhbnNmb3JtOwp9CgppbXBvcnQgX192aXRlX19janNJbXBvcnQyX3JlYWN0X2pzeERldlJ1bnRpbWUgZnJvbSAiL25vZGVfbW9kdWxlcy8udml0ZS9kZXBzL3JlYWN0X2pzeC1kZXYtcnVudGltZS5qcz92PTJlODI1N2Q2IjsgY29uc3QgX2pzeERFViA9IF9fdml0ZV9fY2pzSW1wb3J0Ml9yZWFjdF9qc3hEZXZSdW50aW1lWyJqc3hERVYiXTsKY29uc29sZS5sb2coJ0FwcC5qc3ggMScpIC8vIGZvciBzb3VyY2VtYXAKOwpmdW5jdGlvbiBBcHAoKSB7CiAgICByZXR1cm4gLyojX19QVVJFX18qLyBfanN4REVWKCJkaXYiLCB7CiAgICAgICAgb25DbGljazogKCk9PnsKICAgICAgICAgICAgdGhyb3cgbmV3IEVycm9yKCdiYXInKTsKICAgICAgICB9LAogICAgICAgIGNoaWxkcmVuOiAiZm9vIgogICAgfSwgdm9pZCAwLCBmYWxzZSwgewogICAgICAgIGZpbGVOYW1lOiAiL1VzZXJzL2FybmF1ZC9naXQvdml0ZS1wbHVnaW4tcmVhY3QvcGxheWdyb3VuZC9yZWFjdC1zb3VyY2VtYXAvQXBwLmpzeCIsCiAgICAgICAgbGluZU51bWJlcjogNCwKICAgICAgICBjb2x1bW5OdW1iZXI6IDUKICAgIH0sIHRoaXMpOwp9Cl9jID0gQXBwOwpjb25zb2xlLmxvZygnQXBwLmpzeCAyJykgLy8gZm9yIHNvdXJjZW1hcAo7CmV4cG9ydCBkZWZhdWx0IEFwcDsKdmFyIF9jOwokUmVmcmVzaFJlZyQoX2MsICJBcHAiKTsKCgppZiAoaW1wb3J0Lm1ldGEuaG90ICYmICFpbldlYldvcmtlcikgewogIHdpbmRvdy4kUmVmcmVzaFJlZyQgPSBwcmV2UmVmcmVzaFJlZzsKICB3aW5kb3cuJFJlZnJlc2hTaWckID0gcHJldlJlZnJlc2hTaWc7Cn0KCgppZiAoaW1wb3J0Lm1ldGEuaG90ICYmICFpbldlYldvcmtlcikgewogIFJlZnJlc2hSdW50aW1lLl9faG1yX2ltcG9ydChpbXBvcnQubWV0YS51cmwpLnRoZW4oKGN1cnJlbnRFeHBvcnRzKSA9PiB7CiAgICBSZWZyZXNoUnVudGltZS5yZWdpc3RlckV4cG9ydHNGb3JSZWFjdFJlZnJlc2goIi9Vc2Vycy9hcm5hdWQvZ2l0L3ZpdGUtcGx1Z2luLXJlYWN0L3BsYXlncm91bmQvcmVhY3Qtc291cmNlbWFwL0FwcC5qc3giLCBjdXJyZW50RXhwb3J0cyk7CiAgICBpbXBvcnQubWV0YS5ob3QuYWNjZXB0KChuZXh0RXhwb3J0cykgPT4gewogICAgICBpZiAoIW5leHRFeHBvcnRzKSByZXR1cm47CiAgICAgIGNvbnN0IGludmFsaWRhdGVNZXNzYWdlID0gUmVmcmVzaFJ1bnRpbWUudmFsaWRhdGVSZWZyZXNoQm91bmRhcnlBbmRFbnF1ZXVlVXBkYXRlKCIvVXNlcnMvYXJuYXVkL2dpdC92aXRlLXBsdWdpbi1yZWFjdC9wbGF5Z3JvdW5kL3JlYWN0LXNvdXJjZW1hcC9BcHAuanN4IiwgY3VycmVudEV4cG9ydHMsIG5leHRFeHBvcnRzKTsKICAgICAgaWYgKGludmFsaWRhdGVNZXNzYWdlKSBpbXBvcnQubWV0YS5ob3QuaW52YWxpZGF0ZShpbnZhbGlkYXRlTWVzc2FnZSk7CiAgICB9KTsKICB9KTsKfQoKLy8jIHNvdXJjZU1hcHBpbmdVUkw9ZGF0YTphcHBsaWNhdGlvbi9qc29uO2Jhc2U2NCxleUoyWlhKemFXOXVJam96TENKemIzVnlZMlZ6SWpwYklrRndjQzVxYzNnaVhTd2ljMjkxY21ObGMwTnZiblJsYm5RaU9sc2lZMjl1YzI5c1pTNXNiMmNvSjBGd2NDNXFjM2dnTVNjcElDOHZJR1p2Y2lCemIzVnlZMlZ0WVhCY2JtWjFibU4wYVc5dUlFRndjQ2dwSUh0Y2JpQWdjbVYwZFhKdUlDaGNiaUFnSUNBOFpHbDJYRzRnSUNBZ0lDQnZia05zYVdOclBYc29LU0E5UGlCN1hHNGdJQ0FnSUNBZ0lIUm9jbTkzSUc1bGR5QkZjbkp2Y2lnblltRnlKeWxjYmlBZ0lDQWdJSDE5WEc0Z0lDQWdQbHh1SUNBZ0lDQWdabTl2WEc0Z0lDQWdQQzlrYVhZK1hHNGdJQ2xjYm4xY2JseHVZMjl1YzI5c1pTNXNiMmNvSjBGd2NDNXFjM2dnTWljcElDOHZJR1p2Y2lCemIzVnlZMlZ0WVhCY2JseHVaWGh3YjNKMElHUmxabUYxYkhRZ1FYQndYRzRpWFN3aWJtRnRaWE1pT2xzaVkyOXVjMjlzWlNJc0lteHZaeUlzSWtGd2NDSXNJbVJwZGlJc0ltOXVRMnhwWTJzaUxDSkZjbkp2Y2lKZExDSnRZWEJ3YVc1bmN5STZJanM3T3pzN096czdPenM3T3pzN096czdPenM3UVVGQlFVRXNVVUZCVVVNc1IwRkJSeXhEUVVGRExHRkJRV0VzWjBKQlFXZENPenRCUVVONlF5eFRRVUZUUXp0SlFVTlFMSEZDUVVORkxGRkJRVU5ETzFGQlEwTkRMRk5CUVZNN1dVRkRVQ3hOUVVGTkxFbEJRVWxETEUxQlFVMDdVVUZEYkVJN2EwSkJRMFE3T3pzN096dEJRVWxNTzB0QlZsTklPMEZCV1ZSR0xGRkJRVkZETEVkQlFVY3NRMEZCUXl4aFFVRmhMR2RDUVVGblFqczdRVUZGZWtNc1pVRkJaVU1zU1VGQlJ5Sjk1NzYAeyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIkFwcC5qc3giXSwic291cmNlc0NvbnRlbnQiOlsiY29uc29sZS5sb2coJ0FwcC5qc3ggMScpIC8vIGZvciBzb3VyY2VtYXBcbmZ1bmN0aW9uIEFwcCgpIHtcbiAgcmV0dXJuIChcbiAgICA8ZGl2XG4gICAgICBvbkNsaWNrPXsoKSA9PiB7XG4gICAgICAgIHRocm93IG5ldyBFcnJvcignYmFyJylcbiAgICAgIH19XG4gICAgPlxuICAgICAgZm9vXG4gICAgPC9kaXY+XG4gIClcbn1cblxuY29uc29sZS5sb2coJ0FwcC5qc3ggMicpIC8vIGZvciBzb3VyY2VtYXBcblxuZXhwb3J0IGRlZmF1bHQgQXBwXG4iXSwibmFtZXMiOlsiY29uc29sZSIsImxvZyIsIkFwcCIsImRpdiIsIm9uQ2xpY2siLCJFcnJvciJdLCJtYXBwaW5ncyI6Ijs7Ozs7Ozs7Ozs7Ozs7Ozs7Ozs7QUFBQUEsUUFBUUMsR0FBRyxDQUFDLGFBQWEsZ0JBQWdCOztBQUN6QyxTQUFTQztJQUNQLHFCQUNFLFFBQUNDO1FBQ0NDLFNBQVM7WUFDUCxNQUFNLElBQUlDLE1BQU07UUFDbEI7a0JBQ0Q7Ozs7OztBQUlMO0tBVlNIO0FBWVRGLFFBQVFDLEdBQUcsQ0FBQyxhQUFhLGdCQUFnQjs7QUFFekMsZUFBZUMsSUFBRyJ9) it map a bit more (onClick) but fails the children is off

I'm still ok for going the full first line inline way, it feels hacky for me but probably only people like me look at generated code, the same way normal people never look at the URL 😅

sapphi-red: > ```js
> 'AAAA;'.repeat(code.split('\n').length - 1),
> ```

I think this works similarly to `hires: false` and probably that's the reason the sourcemap is a bit broken when combined with babel.

I prefer to go with the current way for now.

ArnaudBarre: Nope the soucemap is perfectly valid, it just says the line didn't change which is true. When combined it works well with the OXC sourcemap
On the babel plugin the sourcemap is not working great but that's nothing to do with `'AAAA;'.repeat(code.split('\n').length - 1)` which I added only on the OXC plugin

sapphi-red: > Nope the soucemap is perfectly valid, it just says the line didn't change which is true.

`'AAAA;'.repeat(code.split('\n').length - 1)` only maps the first character of each lines (which is the same behavior with `hires: false`). If I remember correctly, the position of `console.log` will be at the first character of the line rather than the position next to `console.log`. This is not that important as we have multiple places doing similar to it (it's better if we can avoid them though). But in addition to that, it causes troubles when combined with other sourcemaps.
https://x.com/bluwyoo/status/1674036158714769408 (I didn't find any examples, only found this post 😅)

sapphi-red: What we want is a range mapping (a proposal exists: https://github.com/tc39/ecma426/blob/main/proposals/range-mappings.md), but currently sourcemaps only supports character-by-character mapping.

ArnaudBarre: Ok thanks I see and I understand now why people wants a new source map spec 😅
That's sad that such a common usecase is not easy to do 😢

packages/common/refresh-utils.ts
```diff
@@ -14,20 +14,30 @@ window.$RefreshSig$ = () => (type) => type;`
 export const getPreambleCode = (base: string): string =>
   preambleCode.replace('__BASE__', base)
 
-export function addRefreshWrapper<M extends { mappings: string } | undefined>(
+export const avoidSourceMapOption = Symbol()
+
+export function addRefreshWrapper<M extends { mappings: string }>(
   code: string,
-  map: M | string,
+  map: M | string | undefined | typeof avoidSourceMapOption,
```
ArnaudBarre: I think we don't need `undefined`?

packages/plugin-react-oxc/src/index.ts
```diff
@@ -0,0 +1,158 @@
...
+      filter: { id: exactRegex(runtimePublicPath) },
+      handler(id) {
+        return id
+      },
+    },
+    load: {
+      filter: { id: exactRegex(runtimePublicPath) },
+      handler(_id) {
+        return readFileSync(
+          join(_dirname, 'refresh-runtime.js'),
```
ArnaudBarre, line 92: Note related to this PR because the issue also exist in the babel plugin but this is not working in dev.
I not sure yet what is the best way to handle this (I though about adding `src/refresh-runtime.js` to gitignore and having a copy step in the dev script, but not 100% convinced)

ArnaudBarre (approved):

packages/plugin-react-oxc/src/index.ts
sapphi-red: I'll take a look if I can fix this in a separate PR 👍

Issue #11, opened 2021-09-14
[plugin-react-refresh] auto page reloading is broken when a jsx file contain multi-components


kljzndx:
### Describe the bug

For details. Please view the video

![video](https://s3.bmp.ovh/imgs/2021/09/f0490350a25738b6.gif)

### Reproduction

https://github.com/kljzndx/Test-Demos/tree/vite-react-bug

### System Info

```shell
System:
    OS: Linux 5.10 Debian GNU/Linux 10 (buster) 10 (buster)
    CPU: (4) arm64 unknown
    Memory: 1.52 GB / 3.71 GB
    Container: Yes
    Shell: 5.7.1 - /usr/bin/zsh
  Binaries:
    Node: 14.16.0 - /usr/local/bin/node
    Yarn: 1.22.11 - /usr/local/bin/yarn
    npm: 7.23.0 - /usr/local/bin/npm
  Browsers:
    Chromium: 89.0.4389.114 # Raspberry pi
    Edge: 93.0.961.47 # local PC
```

### Used Package Manager

yarn

### Validations

- [X] Follow our [Code of Conduct](https://github.com/vitejs/vite/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [Contributing Guidelines](https://github.com/vitejs/vite/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://vitejs.dev/guide).
- [X] Check that there isn't [already an issue](https://github.com/vitejs/vite/issues) that reports the same bug to avoid creating a duplicate.
- [X] Make sure this is a Vite issue and not a framework-specific issue. For example, if it's a Vue SFC related bug, it should likely be reported to https://github.com/vuejs/vue-next instead.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/vitejs/vite/discussions) or join our [Discord Chat Server](https://chat.vitejs.dev/).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

IanVS:
React fast refresh does not work with `class` components, only functional components.  There's an [issue vitejs/vite-plugin-react#8](https://github.com/vitejs/vite-plugin-react/issues/8) open to include class components.

kljzndx:
好吧，看来是我误解了。但是为什么文档里面只字未提这件事？
--------------Google Translate------------
[...]

PR #443, opened 2025-04-07
refactor: extract common code in common package

sapphi-red:
### Description

Extracts common code from plugin-react and plugin-react-swc into `packages/common`. This package will be bundled to both plugins so the number of dependencies are still the same.

Behavior changes in each commit (due to misalignment of between plugins):

- 707fd5d01e90be3220b712933505085dccbd58dc: none
- 20d67edf1204f540d984c880e9380e4b9e825afe
  - [plugin-react-swc] The `Could not Fast Refresh` error now points to `https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react-swc#consistent-components-exports` instead of `https://github.com/vitejs/vite-plugin-react-swc#consistent-components-exports`
  - [plugin-react] Properties other than `injectIntoGlobalHook` are removed from the default export object. I treated properties other than `injectIntoGlobalHook` is a internal thing, but it may treated as a breaking change.
- c78fd216b2e09826b66d31c90b05e93188e3e3b0
  - [plugin-react] `window.__vite_plugin_react_preamble_installed__` is removed. This should not change any behavior.
  - ~~[plugin-react] `viteReact.preambleCode` is deprecated and `viteReact.getPreambleCode` is newly exposed~~ reverted in 8d56222013c3cbb4bfe05e5a17a3211088a30137
- 2ea890ad7f7dc2ef323f41a62d61cea37b84a65a
  - [plugin-react-swc] `import.meta.hot && !inWebWorker` check is added.
  - [plugin-react-swc] Cannot detect preamble error message is slightly changed.
  - [plugin-react] `See https://github.com/vitejs/vite-plugin-react/pull/11#discussion_r430879201` was removed from the cannot detect preamble error as it was stale.
- d144462d2b1d1b4ef6a137c642b83edc955e65bd: none


ArnaudBarre (commented):
Thanks for this, this will make the new plugin so small !

Two notes for me:
- I don't think going for the preambleCode->getPreambleCode deprecation is worth the added benefit for us or users.
- I was happy without the `import.meta.hot && !inWebWorker` check in SWC because for me this checks only hide bundling issues. I'm afraid that if one day we implement HMR on webworker, we will break people that use to bundle random things in their worker. But given that most users as Babel users, it make sense to align in this direction

packages/plugin-react-swc/src/index.ts
```diff
@@ -151,43 +147,13 @@ const react = (_options?: Options): PluginOption[] => {
         if (!result) return
         if (!refresh) return result
 
-        const hasRefresh = refreshContentRE.test(result.code)
-        if (!hasRefresh && !reactCompRE.test(result.code)) return result
-
         const sourceMap: SourceMapPayload = JSON.parse(result.map!)
```
ArnaudBarre: There is a slight behaviour change here: sourceMap is parsed for non 'refreshed' files

sapphi-red: Ah, I thought about that and moved the `JSON.parse` into `addRefreshWrapper` and then forgot about removing it here.

sapphi-red:
> * I don't think going for the preambleCode->getPreambleCode deprecation is worth the added benefit for us or users.

That's true. I'll remove this change.

sapphi-red:
I've updated 👍

ArnaudBarre (approved):

PR #445, opened 2025-04-09
chore: add type field to plugin-react package

sapphi-red:
### Description

To fix a publint warning.
https://publint.dev/rules#use_type


sapphi-red:
Merging as this is a simple change and shouldn't cause any problems.

PR #451, opened 2025-04-14
feat: add support for rolldown-vite

sapphi-red:
### Description

Moved the patch from https://github.com/vitejs/vite-ecosystem-ci/blob/rolldown-vite/README-temp.md.
This was tested with ecosystem-ci.


packages/plugin-react/src/index.ts
```diff
@@ -203,7 +203,14 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
         ])
       }
 
-      if (opts.jsxRuntime === 'classic' && isJSX) {
+      if (
+        opts.jsxRuntime === 'classic' &&
+        isJSX &&
+        // OXC injects self and source so these plugins are not needed for rolldown-vite
```
ArnaudBarre, line 209: Why is that? Does OXC run before Babel with rolldown-vite?
Technically it works without in default Vite but the source info is off which makes info in React devtools wrong (and breaks https://github.com/ArnaudBarre/vite-plugin-react-click-to-component)

sapphi-red: Oxc runs after this plugin and injects source and self. Oxc rejects an JSX with self and source with `Duplicate __self prop found.` error.

sapphi-red: In babel, the error only happens with automatic runtime. It looks like Oxc should align the behavior to that.
Would it be fine to add this change for now to unblock the error and revert it later?

sapphi-red: Ah, probably we should set `development: false`.

sapphi-red: Hmm, it seems it still errors with `development: false`.

sapphi-red: Made an issue: https://github.com/oxc-project/oxc/issues/10391

ArnaudBarre: Ok I see. Yeah we can merge this as this is just a small DX issue

ArnaudBarre:
Can you add a line to the changelog?

sapphi-red:
I added a changelog 👍

ArnaudBarre (approved):

ArnaudBarre:
You will need to rebase for the changelog to not be in the beta version!

PR #456, opened 2025-04-15
chore(deps): update react compiler version

sapphi-red:
### Description

update the react compiler version to latest beta. also remove them from renovate ignoreDeps.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -30,9 +30,5 @@
     "kill-port", // `kill-port:^2.0.0 has perf issues (#8392)
 
     "prettier", // waiting for stable choice on ternaries
-
-    // renovate doesn't properly handle x.x.x-experimental-yyy-zzz version schema
-    "react-compiler-runtime",
-    "babel-plugin-react-compiler",
   ],
 }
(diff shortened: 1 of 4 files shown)


.github/renovate.json5
```diff
@@ -30,9 +30,5 @@
     "kill-port", // `kill-port:^2.0.0 has perf issues (#8392)
 
     "prettier", // waiting for stable choice on ternaries
-
-    // renovate doesn't properly handle x.x.x-experimental-yyy-zzz version schema
```
ArnaudBarre, line 34: We need to keep this as they are still putting the hash before the date 🤦‍♂️
(otherwise it can update backwards when a new hash is "before" the current one)

sapphi-red: Ah, I see. I added the followTag option instead now.
https://docs.renovatebot.com/configuration-options/#followtag
It probably works as it has the `latest` tag.
https://www.npmjs.com/package/react-compiler-runtime?activeTab=versions

ArnaudBarre: Ah good! Let's be vigilant on the next auto update from the bot

ArnaudBarre (approved):

PR #467, opened 2025-04-26
fix(react): inject `__self` and `__source` by babel for rolldown-vite

sapphi-red:
### Description

fix for https://github.com/vitejs/vite-plugin-react/pull/451#discussion_r2041570957

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -131,7 +131,8 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
               jsx: {
                 runtime: 'classic',
                 // disable __self and __source injection even in dev
-                // as this plugin injects them by babel
+                // as this plugin injects them by babel and oxc will throw
+                //  if development is enabled and those properties are already present
                 development: false,
               },
             },
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -4,6 +4,7 @@ import { readFileSync } from 'node:fs'
 import type * as babelCore from '@babel/core'
 import type { ParserOptions, TransformOptions } from '@babel/core'
 import { createFilter } from 'vite'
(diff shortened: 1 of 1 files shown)


packages/plugin-react/src/index.ts
```diff
@@ -124,10 +125,23 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
     enforce: 'pre',
     config() {
       if (opts.jsxRuntime === 'classic') {
-        return {
-          esbuild: {
-            jsx: 'transform',
-          },
+        if ('rolldownVersion' in vite) {
```
sapphi-red, line 128: `this.meta` does not exist in `config` hook (yet).

packages/plugin-react/src/index.ts
```diff
@@ -124,10 +125,23 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
...
+          return {
+            oxc: {
+              jsx: {
+                runtime: 'classic',
+                // disable __self and __source injection even in dev
+                // as this plugin injects them by babel
```
ArnaudBarre: Adding this because the initial reason was not obvious (compare to esbuild that ignores it)
```suggestion
                // as this plugin injects them by babel and oxc will throw
                //  if development is enabled and those properties are already present
```

ArnaudBarre (approved):

packages/plugin-react/src/index.ts
```diff
@@ -131,7 +131,8 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
               jsx: {
                 runtime: 'classic',
                 // disable __self and __source injection even in dev
-                // as this plugin injects them by babel
+                // as this plugin injects them by babel and oxc will throw
+                //  if development is enabled and those properties are already present
```
```suggestion
                // if development is enabled and those properties are already present
```

ArnaudBarre (approved):

PR #470, opened 2025-04-28
feat: add filter

sapphi-red:
### Description

Adds [`filter`](https://rolldown.rs/guide/plugin-development#plugin-hook-filters) so that it makes these plugin performant when used in rolldown-vite.

### DIFF
--- a/packages/plugin-react-swc/src/index.ts
+++ b/packages/plugin-react-swc/src/index.ts
@@ -142,13 +142,17 @@ const react = (_options?: Options): PluginOption[] => {
           )
         }
       },
-      transformIndexHtml: (_, config) => [
-        {
-          tag: 'script',
-          attrs: { type: 'module' },
-          children: getPreambleCode(config.server!.config.base),
-        },
-      ],
+      transformIndexHtml: (_, config) => {
+        if (!hmrDisabled) {
+          return [
+            {
+              tag: 'script',
+              attrs: { type: 'module' },
+              children: getPreambleCode(config.server!.config.base),
+            },
+          ]
+        }
+      },
       async transform(code, _id, transformOptions) {
         const id = _id.split('?')[0]
         const refresh = !transformOptions?.ssr && !hmrDisabled
--- a/packages/plugin-react-swc/src/index.ts
+++ b/packages/plugin-react-swc/src/index.ts
@@ -14,6 +14,7 @@ import {
 import type { PluginOption } from 'vite'
 import {
   addRefreshWrapper,
+  exactRegex,
   getPreambleCode,
   runtimePublicPath,
   silenceUseClientWarning,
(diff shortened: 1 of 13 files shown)


packages/plugin-react-swc/src/index.ts
```diff
@@ -96,14 +97,23 @@ const react = (_options?: Options): PluginOption[] => {
...
-      load: (id) =>
-        id === runtimePublicPath
-          ? readFileSync(join(_dirname, 'refresh-runtime.js'), 'utf-8').replace(
-              /__README_URL__/g,
-              'https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react-swc',
-            )
-          : undefined,
+      resolveId: {
+        filter: { id: exactRegex(runtimePublicPath) },
+        handler: (id) => (id === runtimePublicPath ? id : undefined),
```
sapphi-red, line 102: We still need to keep the `if` statements to keep the compat for Vite older than v6.3.

packages/plugin-react/src/index.ts
```diff
@@ -181,114 +185,128 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
...
-            await loadPlugin('@babel/plugin-transform-react-jsx-self'),
-            await loadPlugin('@babel/plugin-transform-react-jsx-source'),
+    transform: {
+      filter: {
+        id: {
+          include: ensureArray(include).map(matchWithQuery),
+          exclude: [
+            ...(exclude ? ensureArray(exclude).map(matchWithQuery) : []),
+            /\/node_modules\//,
+          ],
```
ArnaudBarre: This filter conversion and the two utils feels a bit tricky IMO.
Could this be upstream into the Vite `createFilter` function so that it can be reused by other plugins?

packages/plugin-react/src/index.ts
```diff
@@ -181,114 +185,128 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
         // we only create static option in this case and re-create them
         // each time otherwise
         staticBabelOptions = createBabelOptions(opts.babel)
+
+        if (
+          canSkipBabel(staticBabelOptions.plugins, staticBabelOptions) &&
+          skipFastRefresh &&
+          isProduction
```
ArnaudBarre: `isProduction` implies `skipFastRefresh` for now so we can skip the second check but it doesn't hurt so up to you

packages/plugin-react/src/index.ts
sapphi-red: If we add a util function to Vite, we need to bump the required peer dep in this plugin (and the other plugins that wants to use that) which is a breaking change. I think if we do that we probably should add it to a separate package.

I'll consider if we can have `@rolldown/pluginutils` (since I sent a PR to `@rollup/pluginutils` that adds some utils functions and it's taking some time to be merged).

ArnaudBarre: This could be added and then we could do `import * as Vite from 'vite'` and check for the export name existence so that people using the latest version have the speedup (which is expecting for people running rolldown vite)

packages/plugin-react/src/index.ts
sapphi-red: Actually, I think this is needed. For example, when `config.command === 'build' && !isProduction` is true (`NODE_ENV=development vite build`), this if block should not be executed. Otherwise `@babel/plugin-transform-react-jsx-self` / `@babel/plugin-transform-react-jsx-source` won't run.

packages/plugin-react/src/index.ts
sapphi-red: I've updated the code to use `@rolldown/pluginutils`. `ensureArray` is still there because we need to combine with `/\/node_modules\//`, but I think it is now simple enough.

ArnaudBarre (approved):

sapphi-red:
Just noticed that I forgot to add the changelog. Added 👍

PR #489, opened 2025-06-03
fix: use `optimizeDeps.rollupOptions` for rolldown-vite

sapphi-red:
### Description

So that the warning is not output.

refs https://github.com/vitejs/rolldown-vite/issues/207


ArnaudBarre (approved):

sapphi-red:
I noticed that I forgot the changelog. I added that.

PR #490, opened 2025-06-03
chore(readme): add plugin-react-oxc to the table

TheAlexLichter:
The oxc package is missing from the list. In case it has been deliberately omitted, feel free to close the PR 👍🏻


ArnaudBarre (approved):
Yes we should add it, Evan agreed we should encourage people to try it with rolldown-vite!

TheAlexLichter:
> Yes we should add it, Evan agreed we should encourage people to try it with rolldown-vite!

[Another PR](https://github.com/vitejs/vite-plugin-react/pull/491) to help with that even more comes any minute 😁

PR #491, opened 2025-06-03
feat: suggest `vite-plugin-react-oxc` when using `rolldown-vite`

TheAlexLichter:
### Description

This PR introduces a log (warn) when using `rolldown-vite` and either:

* `vite-plugin-react`,
* or `vite-plugin-react-swc` (without any `swc` plugins set)

It recommends using `vite-plugin-react-oxc` and links to the Rolldown migration guide.

I'm planning on adding info regarding `vite-plugin-react-oxc` to it in another PR

### DIFF
--- a/packages/plugin-react-swc/README.md
+++ b/packages/plugin-react-swc/README.md
@@ -119,7 +119,7 @@ react({
 
 ### disableOxcRecommendation
 
-If set, disables the recommendation to use `vite-plugin-react-oxc` (which is shown when `rolldown-vite` is detected and neither `swc` plugins are used nor the `swc` options are mutated).
+If set, disables the recommendation to use `@vitejs/plugin-react-oxc` (which is shown when `rolldown-vite` is detected and neither `swc` plugins are used nor the `swc` options are mutated).
 
 ```ts
 react({ disableOxcRecommendation: true })
--- a/packages/plugin-react-swc/README.md
+++ b/packages/plugin-react-swc/README.md
@@ -117,6 +117,14 @@ react({
 })
 ```
 
+### disableOxcRecommendation
+
+If set, disables the recommendation to use `vite-plugin-react-oxc` (which is shown when `rolldown-vite` is detected and neither `swc` plugins are used nor the `swc` options are mutated).
+
+```ts
+react({ disableOxcRecommendation: true })
+```
+
 ## Consistent components exports
 
 For React refresh to work correctly, your file should only export React components. The best explanation I've read is the one from the [Gatsby docs](https://www.gatsbyjs.com/docs/reference/local-development/fast-refresh/#how-it-works).
--- a/packages/plugin-react-swc/src/index.ts
+++ b/packages/plugin-react-swc/src/index.ts
@@ -153,12 +153,13 @@ const react = (_options?: Options): PluginOption[] => {
          * Suggest to use vite-plugin-react-oxc if `rolldown-vite` is used and:
          * No swc plugins are set
          * mutateSwcOptions is not set
+         * It is not disabled by the user
          */
         if (
           'rolldownVersion' in vite &&
-          !options.disableOxcRecommendation &&
+          !options.plugins &&
           !options.useAtYourOwnRisk_mutateSwcOptions &&
-          !options.plugins
+          !options.disableOxcRecommendation
         ) {
           config.logger.warn(
             '[vite:react-swc] We recommend switching to `vite-plugin-react-oxc` for improved performance as no swc plugins are used. More information at https://vite.dev/rolldown',
--- a/packages/plugin-react-swc/src/index.ts
+++ b/packages/plugin-react-swc/src/index.ts
@@ -78,7 +78,7 @@ type Options = {
   useAtYourOwnRisk_mutateSwcOptions?: (options: SWCOptions) => void
 
   /**
-   * If set, disables the recommendation to use `@vitejs/plugin-react-oxc`
+   * If set, disables the recommendation to use `vite-plugin-react-oxc`
    */
   disableOxcRecommendation?: boolean
 }
@@ -160,7 +160,7 @@ const react = (_options?: Options): PluginOption[] => {
           !options.useAtYourOwnRisk_mutateSwcOptions &&
           !options.plugins
         ) {
-          console.warn(
+          config.logger.warn(
             '[vite:react-swc] We recommend switching to `vite-plugin-react-oxc` for improved performance as no swc plugins are used. More information at https://vite.dev/rolldown',
           )
         }
--- a/packages/plugin-react-swc/src/index.ts
+++ b/packages/plugin-react-swc/src/index.ts
@@ -76,6 +76,11 @@ type Options = {
    * feature doesn't work is not fun, so we won't provide support for it, hence the name `useAtYourOwnRisk`
    */
   useAtYourOwnRisk_mutateSwcOptions?: (options: SWCOptions) => void
+
+  /**
+   * If set, disables the recommendation to use `vite-plugin-react-oxc`
+   */
+  disableOxcRecommendation?: boolean
 }
 
 const react = (_options?: Options): PluginOption[] => {
@@ -91,6 +96,7 @@ const react = (_options?: Options): PluginOption[] => {
     reactRefreshHost: _options?.reactRefreshHost,
     useAtYourOwnRisk_mutateSwcOptions:
       _options?.useAtYourOwnRisk_mutateSwcOptions,
+    disableOxcRecommendation: _options?.disableOxcRecommendation,
   }
 
   return [
@@ -143,12 +149,14 @@ const react = (_options?: Options): PluginOption[] => {
           )
         }
 
-        /* Suggest to use vite-plugin-react-oxc if `rolldown-vite` is used and:
+        /*
+         * Suggest to use vite-plugin-react-oxc if `rolldown-vite` is used and:
          * No swc plugins are set
          * mutateSwcOptions is not set
          */
         if (
           'rolldownVersion' in vite &&
+          !options.disableOxcRecommendation &&
           !options.useAtYourOwnRisk_mutateSwcOptions &&
           !options.plugins
         ) {
--- a/packages/plugin-react-swc/src/index.ts
+++ b/packages/plugin-react-swc/src/index.ts
@@ -78,7 +78,7 @@ type Options = {
   useAtYourOwnRisk_mutateSwcOptions?: (options: SWCOptions) => void
 
   /**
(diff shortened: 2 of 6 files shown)


packages/plugin-react-swc/src/index.ts
```diff
@@ -141,6 +142,13 @@ const react = (_options?: Options): PluginOption[] => {
             '[vite:react-swc] The MDX plugin should be placed before this plugin',
           )
         }
+
+        // Suggest to use vite-plugin-react-oxc if `rolldown-vite` is used and no swc plugins are set
+        if ('rolldownVersion' in vite && !config.plugins.length) {
```
ArnaudBarre: We should check for useAtYourOwnRisk_mutateSwcOptions also

packages/plugin-react-swc/src/index.ts
```diff
@@ -141,6 +142,13 @@ const react = (_options?: Options): PluginOption[] => {
...
+          console.warn(
+            '[vite:react-swc] We recommend switching to `vite-plugin-react-oxc` for improved performance as no swc plugins are used. More information at https://vite.dev/rolldown',
```
ArnaudBarre: I'm not 100% sure about performances. In dev mode, transforming ts is very rarely the bottleneck and in build mode esbuild is used and also very rarely the bottleneck (it's source map generation for me).
But the nice benefit is one less dependency and better integrated with Vite

packages/plugin-react/src/index.ts
```diff
@@ -131,6 +131,12 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
     name: 'vite:react-babel',
     enforce: 'pre',
     config() {
+      if ('rolldownVersion' in vite) {
```
ArnaudBarre: We should check for babel plugins, some people may depend on some weird legacy things that will not be supported by oxc

ArnaudBarre (commented):
I think that's a good way to make people aware of it, but maybe we should provide a way to disable the warning, but one extra config field for this feels too much. I'm curious of the opinion of other team member @patak-dev @sapphi-red @hi-ogawa

TheAlexLichter:
Thanks for your notes!

> maybe we should provide a way to disable the warning, but one extra config field for this feels too much.

I had the same thoughts and didn't come up with a better solution so far. But maybe we could use an env var instead to avoid "one more config option" and cater to the needs of the few who want to disable it? 🤔

packages/plugin-react-swc/src/index.ts
TheAlexLichter: Will check that, thanks!

packages/plugin-react-swc/src/index.ts
TheAlexLichter: Indeed, I think the big benefit (besides a speed-up during build for some apps) is the unification in terms of dependencies. With `rolldown-vite`, we ideally want to enable people using solely oxc (no esbuild, as mentioned in the migration guide and also in [the blog post](https://voidzero.dev/posts/announcing-rolldown-vite#esbuild-as-optional-dependency)).

packages/plugin-react/src/index.ts
TheAlexLichter: Will do 👍

patak-cat:
> I think that's a good way to make people aware of it, but maybe we should provide a way to disable the warning, but one extra config field for this feels too much. I'm curious of the opinion of other team member @patak-dev @sapphi-red @hi-ogawa

I agree that there should be a way to disable the warning. If the idea is making vite-plugin-react-oxc the default, I think it is ok to have an option in the other plugins config to disable it.

packages/plugin-react/src/index.ts
TheAlexLichter: I've checked for the whole `babel` option. Do you think that is good enough or should I check for `function` _or_ `babel.plugins` being populated?

TheAlexLichter:
Added `disableOxcRecommendation` which is not set by default to both plugins 👍🏻

packages/plugin-react-swc/README.md
```diff
@@ -117,6 +117,14 @@ react({
 })
 ```
 
+### disableOxcRecommendation
+
+If set, disables the recommendation to use `vite-plugin-react-oxc` (which is shown when `rolldown-vite` is detected and neither `swc` plugins are used nor the `swc` options are mutated).
```
sapphi-red: It's published as `@vitejs/plugin-react-oxc`.
```suggestion
If set, disables the recommendation to use `@vitejs/plugin-react-oxc` (which is shown when `rolldown-vite` is detected and neither `swc` plugins are used nor the `swc` options are mutated).
```
(other places as well)

packages/plugin-react/src/index.ts
sapphi-red: I think we also need to skip if there's any plugin that has `api.reactBabel` as that may update the babel option.
https://github.com/vitejs/vite-plugin-react/blob/7679e49bf4433afe13e7df91a0f2e498720d29db/packages/plugin-react/src/index.ts#L189-L191

packages/plugin-react-swc/src/index.ts
```diff
@@ -141,6 +148,22 @@ const react = (_options?: Options): PluginOption[] => {
...
+         * No swc plugins are set
+         * mutateSwcOptions is not set
+         */
+        if (
+          'rolldownVersion' in vite &&
+          !options.disableOxcRecommendation &&
+          !options.useAtYourOwnRisk_mutateSwcOptions &&
+          !options.plugins
+        ) {
+          console.warn(
```
sapphi-red: I think we should use `config.logger.warn`.

sapphi-red:
> > maybe we should provide a way to disable the warning, but one extra config field for this feels too much.
> 
> I had the same thoughts and didn't come up with a better solution so far. But maybe we could use an env var instead to avoid "one more config option" and cater to the needs of the few who want to disable it? 🤔

This isn't directly related, but implementing https://github.com/vitejs/vite/issues/19937 in the future could help users filter out certain warnings more easily.

packages/plugin-react/src/index.ts
TheAlexLichter: Checked for hooks.length 👍

packages/plugin-react-swc/src/index.ts
```diff
@@ -76,6 +76,11 @@ type Options = {
    * feature doesn't work is not fun, so we won't provide support for it, hence the name `useAtYourOwnRisk`
    */
   useAtYourOwnRisk_mutateSwcOptions?: (options: SWCOptions) => void
+
+  /**
+   * If set, disables the recommendation to use `vite-plugin-react-oxc`
```
sapphi-red:
```suggestion
   * If set, disables the recommendation to use `@vitejs/plugin-react-oxc`
```

packages/plugin-react-swc/src/index.ts
```diff
@@ -144,6 +150,23 @@ const react = (_options?: Options): PluginOption[] => {
             '[vite:react-swc] The MDX plugin should be placed before this plugin',
           )
         }
+
+        /*
+         * Suggest to use vite-plugin-react-oxc if `rolldown-vite` is used and:
```
sapphi-red:
```suggestion
         * Suggest to use @vitejs/plugin-react-oxc if `rolldown-vite` is used and:
```

packages/plugin-react-swc/src/index.ts
```diff
@@ -144,6 +150,23 @@ const react = (_options?: Options): PluginOption[] => {
...
+         * It is not disabled by the user
+         */
+        if (
+          'rolldownVersion' in vite &&
+          !options.plugins &&
+          !options.useAtYourOwnRisk_mutateSwcOptions &&
+          !options.disableOxcRecommendation
+        ) {
+          config.logger.warn(
+            '[vite:react-swc] We recommend switching to `vite-plugin-react-oxc` for improved performance as no swc plugins are used. More information at https://vite.dev/rolldown',
```
sapphi-red:
```suggestion
            '[vite:react-swc] We recommend switching to `@vitejs/plugin-react-oxc` for improved performance as no swc plugins are used. More information at https://vite.dev/rolldown',
```

packages/plugin-react/src/index.ts
```diff
@@ -62,6 +62,11 @@ export interface Options {
    * reactRefreshHost: 'http://localhost:3000'
    */
   reactRefreshHost?: string
+
+  /**
+   * If set, disables the recommendation to use `vite-plugin-react-oxc`
```
sapphi-red:
```suggestion
   * If set, disables the recommendation to use `@vitejs/plugin-react-oxc`
```

packages/plugin-react/src/index.ts
```diff
@@ -182,6 +187,23 @@ export default function viteReact(opts: Options = {}): PluginOption[] {
...
+        !opts.disableOxcRecommendation
+      ) {
+        /*
+         * Suggest to use vite-plugin-react-oxc if `rolldown-vite` is used and:
+         * No babel config is set
+         * No other plugin is using the `api.reactBabel` hook
+         * It is not disabled by the user
+         */
+        config.logger.warn(
+          '[vite:react-babel] We recommend switching to `vite-plugin-react-oxc` for improved performance. More information at https://vite.dev/rolldown',
```
sapphi-red:
```suggestion
          '[vite:react-babel] We recommend switching to `@vitejs/plugin-react-oxc` for improved performance. More information at https://vite.dev/rolldown',
```

sapphi-red (commented):
Would you add a changelog entry?

packages/plugin-react-swc/src/index.ts
```diff
@@ -144,6 +150,23 @@ const react = (_options?: Options): PluginOption[] => {
             '[vite:react-swc] The MDX plugin should be placed before this plugin',
           )
         }
+
+        /*
+         * Suggest to use vite-plugin-react-oxc if `rolldown-vite` is used and:
+         * No swc plugins are set
+         * mutateSwcOptions is not set
+         * It is not disabled by the user
+         */
```
ArnaudBarre: This is is pretty clear to me, I don't think we need to maintain twice the list in the future

ArnaudBarre (approved):

packages/plugin-react-swc/src/index.ts
TheAlexLichter: Would you prefer to have it removed then?

ArnaudBarre: Yes personally I would prefer to remove it

TheAlexLichter: Removed both comments 👍

ArnaudBarre (approved):

sapphi-red (approved):

PR #494, opened 2025-06-06
feat: swc cache dir option

RomanHotsiy:
### Description

When using this plugin swc creates an `.swc` older in the cwd. It would be great to allow customizing it especially for meta-frameworks built on top of vite.


RomanHotsiy:
Sorry, wrong repo.

PR #495, opened 2025-06-06
feat: swcCacheDir option

RomanHotsiy:
### Description

When using this plugin swc creates an .swc older in the cwd. It would be great to allow customizing it especially for meta-frameworks built on top of vite.

### DIFF
--- a/packages/plugin-react-swc/README.md
+++ b/packages/plugin-react-swc/README.md
@@ -125,6 +125,14 @@ If set, disables the recommendation to use `@vitejs/plugin-react-oxc` (which is 
 react({ disableOxcRecommendation: true })
 ```
 
+### swcCacheDir
+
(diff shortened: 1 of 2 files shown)


ArnaudBarre (commented):
What value will you use for this option?

I wondering if we should always set it to `node_modules/.cache/swc`

RomanHotsiy:
> What value will you use for this option?

@ArnaudBarre mainly being able to configure it to `node_modules/.cache/swc`. It's fine for projects to add this dir to gitignore but we build meta-framework on top of vite and it's not good UX.

> I wondering if we should always set it to node_modules/.cache/swc

This is a great idea, I've updated the default value. Could you take a look? I would still allow to customize it programmatically for the cases where `node_modules` is not available in cwd (e.g. we run our tool using npx)

PR #518, opened 2025-06-25
feat: add hmr support for compound components

### LINKED ISSUE #493: Add HMR Support for Compound Component Pattern
### Related plugins

- [x] [plugin-react](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react)

### Description

When exporting components as a compound object (e.g. `export const Accordion = { Root, Item }`), React Fast Refresh treats the module as “impure” and falls back to a full reload instead of hot‐swapping the individual subcomponents.

```tsx
// src/Accordion.tsx
import React from 'react'

const Root: React.FC = () => <div>Accordion Root</div>
const Item: React.FC = () => <div>Accordion Item</div>

export const Accordion = { Root, Item }
// → Editing `Root` or `Item` causes a full reload rather than HMR
```

### Suggested solution

Modify the Fast Refresh logic so that any top‐level `export const X = { A, B, … }` is inspected. If **all** properties (`A`, `B`, …) refer to PascalCase functions or classes defined in the same file (i.e. valid React components), treat them as HMR boundaries instead of marking the module as impure.

### Alternative

1. **Barrel export in the same package**

   ```tsx
   // src/components/accordion/components.tsx
   import React from 'react'
   export const Root: React.FC = () => <div>Accordion Root</div>
   export const Item: React.FC = () => <div>Accordion Item</div>

   // src/components/accordion/index.ts
   export * as Accordion from './components'
   ```

  * *Trade-off:* Explicitly re‐exports `Root` and `Item` to enable HMR, but adds extra named exports that clutter IntelliSense.

2. **Separate “micro” package for subcomponents**

```tsx
// packages/accordion/src/components.tsx
import React from 'react'
export const Root: React.FC = () => <div>Accordion Root</div>
export const Item: React.FC = () => <div>Accordion Item</div>

// packages/accordion/src/index.tsx
export * as Accordion from './components'

// apps/app/src/index.tsx
import { Accordion } from "@org/accordion"

return (
   <Accordion.Root>
      <Accordion.Item>...</Accordion.Item>
   </Accordion.Root>
)
```

  * *Trade-off:* While intellisense won't suggested slot components anymore, micro-frontends are not for everyone and they introduce an additional complexity

### Additional context

Compound components (e.g. `<Accordion.Root>`, `<Accordion.Item>`) are a common pattern in UI libraries. Being forced to do a full page reload instead of hot‐swapping slows down development. Enabling HMR for a single object export when all children are valid React components would significantly improve developer experience.

### Validations

Newbie012:
closes https://github.com/vitejs/vite-plugin-react/issues/493


packages/common/refresh-runtime.js
```diff
@@ -545,6 +555,13 @@ function isLikelyComponentType(type) {
   }
 }
 
+function isPlainObject(obj) {
```
ArnaudBarre, line 556: Is there a reason we need this?
`typeof obj === "object" && obj !== null` should be enough no?

ArnaudBarre (commented):
Thanks for working on this and adding the test!
I don't think we need to logic twice, I think this can be done like this:

```js
if (isLikelyComponentType(exportValue)) {
  register(...)
} else if (/* is obj & all keys are component */ ) {
      for (const subKey in exportValue) {
        register(
          exportValue[subKey],
          filename + ' export ' + key + '$' + subKey,
        )
      }
}
```

packages/common/refresh-runtime.js
Newbie012: Maybe, but in your condition, values such as Date, [], Map, etc will return true as well, which is not what we want. Or maybe it's an oversight on my end?

Newbie012:
I'll check and keep you posted

packages/common/refresh-runtime.js
ArnaudBarre: Hum true, it will still be possible to loop over keys but it's cleaner to not do it for these so we can keep this

ArnaudBarre:
My bad the code in isLikelyComponent is needed for the "validateRefreshBoundaryAndEnqueueUpdate" function.

Newbie012:
You're fast! I was just about to comment that I don't think the tests are passing 😅 I reverted back to the original changes.

Newbie012:
Is there something that I can help with in order to merge this change?

ArnaudBarre (approved):

ArnaudBarre:
@Newbie012 I was not very satisfied with the recursive nature of the current code. It would have return true for deeply nested object with only React component, but the runtime parts did not do it recursively. I think one level is enough.
Separating in two functions also avoid registering the top level object of compound components as react component (I think it's a no op but it's cleaner).
I also changed the code to bail out as fast as possible when testing for "compound components"

hi-ogawa (approved):

packages/common/refresh-runtime.js
```diff
@@ -565,6 +580,13 @@ export function registerExportsForReactRefresh(filename, moduleExports) {
       // The register function has an identity check to not register twice the same component,
       // so this is safe to not used the same key here.
       register(exportValue, filename + ' export ' + key)
+    } else if (isCompoundComponent(exportValue)) {
+      for (const subKey in exportValue) {
+        register(
+          exportValue[subKey],
+          filename + ' export ' + key + '$' + subKey,
```
sapphi-red: I think this will conflict when
```js
export const foo = { bar: component1 }
export const foo$bar = component2
```
. I think this can be fixed by picking a character that cannot be used for identifiers like `-` or ` ` instead of `$`.

playground/compound-components/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
```
sapphi-red: [non-blocking] maybe merge this playground to `playground/react` to reduce the number of playgrounds?

playground/compound-components/vite.config.ts
```diff
@@ -0,0 +1,14 @@
+import react from '@vitejs/plugin-react'
+import type { UserConfig } from 'vite'
+
+const config: UserConfig = {
+  server: { port: 8907 /* Should be unique */ },
+  mode: 'development',
```
sapphi-red:
```suggestion
```
I guess this isn't needed. It seems it's set in `playground/react/vite.config.ts` as well though.

sapphi-red (approved):

GuskiS:
Can this be extended to support this syntax?
```tsx
export function Accordion() {
  return <div></div>;
}

Accordion.Root = () => <div>Accordion Root</div>
Accordion.Item = () => <div>Accordion Item</div>
```

PR #562, opened 2025-07-11
fix(rsc): support setups without an SSR environment

jacob-ebey:
### Description

When doing a "SPA" that fetches the RSC payload on boot the plugin currently falls over on production builds due to the lack of a configured rollup input.


hi-ogawa (approved):
Great! I just saw https://github.com/remix-run/react-router-templates/pull/142 and you got no ssr working? 

Technically for this case, we should build through `rsc -> client -> rsc -> client` to cover server references only imported through client. I had an incomplete example for SPA, but just removed it as I thought nobody would care :smile: https://github.com/vitejs/vite-plugin-react/pull/533

I'm happy to add back a minimal example in this repo as a test case. I'll probably merge this one as is and I'll adjust and add minimal demo and test case later. Or if you already have something, please feel free to push them here or new PR. Thanks!

PR #567, opened 2025-07-13
test(rsc): add `navigation` example

grahammendick:
Copied [over from vite-plugins](https://github.com/hi-ogawa/vite-plugins/pull/1137#issuecomment-3066846839).

Hi @hi-ogawa , thank you for working on a vite rsc implementation.

I'm the author of [the Navigation router](https://grahammendick.github.io/navigation/), an rsc framework for react. I'd already written example rsc apps to test out [parcel](https://github.com/grahammendick/navigation/tree/master/NavigationReact/sample/rsc-parcel) and [webpack](https://github.com/grahammendick/navigation/tree/master/NavigationReact/sample/rsc-webpack). So I've started to do the same for your vite implementation. This PR isn't meant to be merged but is a good way to communicate any issues I find.

I cloned your basic example and copied in the files from my working parcel example. When I run `npm run dev` the ssr works but the hydration fails. It fails because the react context isn't populated as expected. It seems like vite has 2 versions of the file - 1 inside the bundled navigation-react.js and one standalone. You can see in the screenshot below it says "from SceneRSCView" instead of "from navigation-react.js". The react context is created from navigation-react.js so it can't be found if it's not accessed the same way. This is just my guess as to what's going wrong.

Screenshot 2025-07-13 at 10 30 57

For some background, The `navigation-react` package has different exports for server and client and has internal client boundaries. The `SceneRSCView`, where the error is happening, is one of these.

Thanks again for all your work.


grahammendick:
@hi-ogawa I can't get npm run build/preview to work, although it was working on vite-plugins and hydration worked in that case.

image

grahammendick:
I'll continue work in vite-plugins PR for now because build/preview works there

hi-ogawa:
Thanks for the PR! The build error was because there was a copy of `examples/basic` specific test case, which should be removed from the example https://github.com/vitejs/vite-plugin-react/blob/44567c350b6b0b887dc404ae7ed81d6eab5515a5/packages/plugin-rsc/examples/basic/vite.config.ts#L91

The dev error is due to Vite's deps optimization. I added `optimizeDeps.include/exclude` in vite.config.ts to fix this. 

It looks like navigation has some issues (both dev and build) and also there's typescript error, which I'm not sure about. Please continue integration and feel free to ask me if anything unclear about Vite RSC. Thanks!

grahammendick:
@hi-ogawa Thank you, that's amazing 🙏

grahammendick:
Thanks @hi-ogawa  for your help. I've finished the navigation example and it works great. It's server-rendered and then all subsequent clicks are fetched using rsc. The code is spit correctly, for example, the `Friends` component isn't loaded until you click 'Show Friends' on the person page. I have some observations that maybe you can help with.

1. There are many requests for the separate js files, rather than bundling them all up into a [single request like parcel does](https://github.com/parcel-bundler/rsc-examples/pull/2#discussion_r1986127645). 
2. I can't use vite's `hydrate` because it makes too many assumptions. So I copied `rscStream` into my client file (parcel exports `rscStream`). Plus, I still had to import vite's `hydrate` even though I'm not using it.
3. Vite assumes that a post request is a server action but I want to fetch rsc over 'post'. I used 'put' for now.
4. The hmr update happens before the modified js files load so the rsc root reloads with the old js.

hi-ogawa:
Thanks for the follow-up! 

> * There are many requests for the separate js files, rather than bundling them all up into a [single request like parcel does](https://github.com/parcel-bundler/rsc-examples/pull/2#discussion_r1986127645).

Yes, this is expected. At the moment, this won't be handled by `@vitejs/plugin-rsc` but it's on user or framework's hand. Custom chunking can be achieved by using underlying bundler option such as Rollup's `manualChunks` or Rolldown's `advancedChunks` https://vite.dev/guide/rolldown.html#manualchunks-to-advancedchunks. I actually haven't tried here, so there might be an issue now, but for example, something like this should work in the future https://github.com/hi-ogawa/vite-plugins/pull/904.

 It's interesting to know that this was the intentional decision by Parcel as I was wondering when I see their behavior in the past. Thanks for the reference!

> * I can't use vite's `hydrate` because it makes too many assumptions. So I copied `rscStream` into my client file (parcel exports `rscStream`). Plus, I still had to import vite's `hydrate` even though I'm not using it.
> * Vite assumes that a post request is a server action but I want to fetch rsc over 'post'. I used 'put' for now.

Right, it wasn't explicitly mentioned here, but `@vitejs/plugin-rsc/extra/...` API is not meant for practical integration. We plan to remove it soon (https://github.com/vitejs/vite-plugin-react/pull/592) and it's expected to write you own integration through `@vitejs/plugin-rsc/{rsc,ssr,browser}` API.

> * The hmr update happens before the modified js files load so the rsc root reloads with the old js.

I'm not familiar with this issue. Can you explain how to reproduce this?

grahammendick:
Thanks for your reply

> it's expected to write you own integration through @vitejs/plugin-rsc/{rsc,ssr,browser} API

Oh yea, I can see the basic example has changed quite a lot since I copied it. I'll have another go using the latest from the basic example. Does that sound ok?

> I'm not familiar with this issue. Can you explain how to reproduce this?

With the Navigation router you don't hard-code Urls. You say what state you want to go to and what data you want to pass and the Navigation router works out the Url using the configuration from the `stateNavigator.ts`. For example, the Navigation router turns `<NavigationLink stateKey="person" navigationData={{ id: 2 }}>` into `<a href="/person/2">`. If you edit the route in `stateNavigator.ts` then the Urls should update automatically. Changing the route from 'person' to 'someone', for example, should update all the Urls in the list of people.
```js
{
  key: 'person',
  route: 'someone/{id}+/{show}',
  defaults: { id: 0, show: false },
  trackCrumbTrail: true,
}
```
But with hmr this update doesn't happen - well, it lags behind. Below are the requests that happen after editing the config in `stateNavigator.ts`. The first request is the new rsc content, fetched by the `HmrProvider` because it received an 'rsc:update' event. The latest `stateNavigator.ts` is only loaded later so I'm assuming that's why the Urls haven't updated correctly.

image

If you update the route back to 'person' in the config again then the Urls do update but to the 'someone' value this time. You see, the hmr change lags behind the config value. I hope that makes sense.

image

hi-ogawa:
> > it's expected to write you own integration through @vitejs/plugin-rsc/{rsc,ssr,browser} API
> 
> Oh yea, I can see the basic example has changed quite a lot since I copied it. I'll have another go using the latest from the basic example. Does that sound ok?

Yes, that would be great :pray: `examples/basic` is more focused on various edge cases and full coverage e2e and it's not meant for the reference template. `examples/starter` is expected to show the full usage of RSC API, so I recommend that one instead.

grahammendick:
Thanks for the help @hi-ogawa. I've followed the starter example and written the Navigation router integration through @vitejs/plugin-rsc/{rsc,ssr,browser} API.

The only issue I have left is [the hmr update](https://github.com/vitejs/vite-plugin-react/pull/567#issuecomment-3098067937). I'll leave this PR open to track it if that's ok with you? Or would you rather I raise a separate issue for it?

hi-ogawa:
What's the expectation of `stateNavigator.ts` modification? If this is meant to manage an entire app state, then browser full reload seems more appropriate and robust behavior. From a quick look, `stateNavigator.ts` doesn't hmr by itself, but parent module `NavigationProvider.tsx` triggers hmr but maybe there's some stale state after that. This is likely a framework side behavior choice and Vite plugin side probably provides enough primitives (and if not, I'm happy to discuss what's needed, but for that, it would be great if you can properly define what's the expected behavior first.)

grahammendick:
Hi @hi-ogawa there was a bug with the Navigation router that prevented Vite HMR from working. I’ve [fixed that](https://github.com/grahammendick/navigation/pull/884) but now I’ve found something odd with Vite HMR that hopefully you can help with.

If you run the latest navigation example and edit the route in the `stateNavigator.tsx` to 'test1{page?}' then all the Urls now update as expected. For example, the Url in the name sort Hyperlink is '/test1?sort=desc'. All good so far.
```js
{
  key: 'people',
  route: 'test1{page?}',
  defaults: { page: 1, sort: 'asc', size: 10 },
},
```
The problem happens when you edit the route a second time to 'test2{page?}', for example. The `NavigationProvider` throws an error because Vite HMR calls it with the latest Url but the old `stateNavigator`. Vite HMR will eventually call it with the updated `stateNavigator` so I can 'fix' it by wrapping in a try/catch. But it's not a proper fix because any client side React state will be lost.

Why does the 2nd Vite HMR update behave so differently to the 1st?

hi-ogawa:
Hey, I really appreciate you continue testing out the HMR capability! 🙏 Unfortunately, I still don't yet have the bandwidth to dig into `navigation` implementation details myself, so I'll need to ask a couple of questions again for follow up:

- Do you have a simple SPA-only example using `navigation` with Vite? How does `stateNavigator` HMR work in that scenario?
- From my testing, when the browser is on `/`, changing other routes multiple times (like `person/{id}+/{show}` → `person-x/{id}+/{show}` → `person-x-y/{id}+/{show}`) seems to work fine. The error appears to be specific to when changing the current route. How does `navigation` handle these two cases differently?

grahammendick:
Hey @hi-ogawa thanks for your response. I've created a separate example called 'hmr' without any navigation dependencies to make the problem clearer. 

The `App` just renders a client `Item` component passing in a key which It gets from 'lookup.ts'. The `Item` passes this key into 'lookup.ts' to get the value which it then displays. If you edit the key in 'lookup.ts' from 'a' to 'b', for example, then the hmr update throws.
```ts
export const key = 'b'
```

The reason for the error is that the server response uses the latest 'lookup.ts' but the client uses the old 'lookup.ts'. So the rsc response has the key 'b' but the client still has the key 'a'.

hi-ogawa:
Great! thanks for the isolated repro :+1: There is indeed something I haven't fully thought through about shared module, which exists both inside server module graph and client module graph

https://github.com/vitejs/vite-plugin-react/blob/aa956279f8f74e4bc04013e803a3c4d062b6a159/packages/plugin-rsc/src/plugin.ts#L432

I would've expected this case being not working at all, so I'm surprised there's some difference between 1st update and 2nd update. This sounds like a concrete bug. Can you create an issue with your repro or just link to this PR?

grahammendick:
@hi-ogawa Thanks, I've raised https://github.com/vitejs/vite-plugin-react/issues/666

grahammendick:
@hi-ogawa Thanks for all your help 🙏 
I'll close this now there's #666 to tracks the hmr issue

PR #707, opened 2025-08-12
chore(rsc): update package.json for starter-cf-single

mashaal:
### Description

Cloning the `starter-cf-single` example with degit, it is missing the `vite` and `rsc-html-stream` and will not run or build.

Adding these and the project works as expected when running:

```sh
npx degit vitejs/vite-plugin-react/packages/plugin-rsc/examples/starter-cf-single starter-cf-single
```


hi-ogawa (commented):
As I didn't include it in README.md, this example isn't technically intended to be "degit"-able but it's only for test case. That said, I don't mind adding two dependencies, so I will merge. Thanks!

PR #798, opened 2025-08-29
perf: remove unnecessary export checks

sapphi-red:
### Description

These checks existed from the first implementation (https://github.com/vitejs/vite/pull/10239), but I think these are not necessary.

This improved the window reload perf a bit. (on the other hand, using `import * as currentExports from 'id'` instead of `RefreshRuntime.__hmr_import(import.meta.url)` degraded the perf, so I'm leaving that as full-bundle mode only)

### DIFF
--- a/packages/common/refresh-runtime.js
+++ b/packages/common/refresh-runtime.js
@@ -653,10 +653,7 @@ export function validateRefreshBoundaryAndEnqueueUpdate(
 
 function predicateOnExport(ignoredExports, moduleExports, predicate) {
   for (const key in moduleExports) {
-    if (key === '__esModule') continue
     if (ignoredExports.includes(key)) continue
-    const desc = Object.getOwnPropertyDescriptor(moduleExports, key)
-    if (desc && desc.get) return key
     if (!predicate(key, moduleExports[key])) return key
   }
   return true


packages/common/refresh-runtime.js
```diff
@@ -653,10 +653,7 @@ export function validateRefreshBoundaryAndEnqueueUpdate(
 
 function predicateOnExport(ignoredExports, moduleExports, predicate) {
   for (const key in moduleExports) {
-    if (key === '__esModule') continue
```
sapphi-red, line 656: Since `moduleExports` is a module namespace, the `__esModule` marker won't exist. If you have `export const __esModule = something`, it should be checked.

packages/common/refresh-runtime.js
```diff
@@ -653,10 +653,7 @@ export function validateRefreshBoundaryAndEnqueueUpdate(
...
     if (ignoredExports.includes(key)) continue
-    const desc = Object.getOwnPropertyDescriptor(moduleExports, key)
-    if (desc && desc.get) return key
```
sapphi-red, line 659: Since `moduleExports` is a module namespace, all properties are not a sideeffectful getter.

hi-ogawa (approved):

ArnaudBarre (approved):

PR #799, opened 2025-08-29
feat(react-compiler): add React Compiler examples

poteto:
This PR adds examples for Vite + React Compiler (JS & TS). These example repos will be used as templates for a new variant I'm proposing to add to create-vite, that allows new Vite apps to be initialized with the compiler ready to go.


yyx990803:
Since we don't really include full examples for the React plugin, I don't think we should add so much code here. It adds the burden of having to sync it with `create-vite` templates in the future.

I think a better addition to this repo specifically is to add documentation on how to use the compiler when you already have `@vitejs/plugin-react` installed.

An additional note here is that in the near future, as rolldown-powered Vite becomes the default version, `@vitejs/plugin-react` will no longer have a Babel dependency. When that happens, usage of React Compiler will likely have to either go through a separate Babel plugin, or there should be a standalone React Compiler plugin that includes Babel as a dependency.

PR #802, opened 2025-08-30
feat(react-swc): set SWC cacheRoot option

ArnaudBarre:
Closes #495

Coming back to this PR, I think using a subpath of the configurable vite cache root option is better than adding an option

cc @RomanHotsiy


sapphi-red (approved):
Sounds like a good way to me 👍

RomanHotsiy:
Thanks

PR #835, opened 2025-09-11
perf: avoid sourcemap for refresh wrapper injection

sapphi-red:
### Description

By changing `$RefreshReg$` / `$RefreshSig$` to a local variable and leveraging the function hoisting, we can avoid the sourcemap generation for the refresh wrapper completely. Also this would reduce the code size, which would also improve the perf a bit.


ArnaudBarre (approved):
Nice

PR #881, opened 2025-09-24
perf(react): use native refresh wrapper plugin in rolldown-vite

sapphi-red:
### Description

Adds some code to use native refresh wrapper plugin in rolldown-vite.

### DIFF
--- a/packages/plugin-react/CHANGELOG.md
+++ b/packages/plugin-react/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


ArnaudBarre:
The perf gain comes from the fact that you skip round-trip for all JSX? files?

sapphi-red:
I believe so (I didn't dig down much though). The other potential reason I can think of is since all files that is processed by JS would be processed serially, even if each process doesn't take that much time, it can be slower than other transforms done on the native side. It also mean that there might be a long queue which would be bad for mutex.

hi-ogawa (approved):

PR #882, opened 2025-09-24
feat(react): full bundle mode compat

sapphi-red:
### Description

Ideally I'd like to avoid having full bundle mode specific code in plugins. But as an initial implementation, I'd like this diff to live in the plugin.

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -117,6 +117,8 @@ export default function viteReact(opts: Options = {}): Plugin[] {
   let isProduction = true
   let projectRoot = process.cwd()
   let skipFastRefresh = true
+  let base: string
+  let isFullBundle = false
   let runPluginOverrides:
     | ((options: ReactBabelOptions, context: ReactBabelHookContext) => void)
     | undefined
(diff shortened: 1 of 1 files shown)


packages/plugin-react/src/index.ts
```diff
@@ -447,15 +454,20 @@ export default function viteReact(opts: Options = {}): Plugin[] {
...
+            {
+              tag: 'script',
+              attrs: { type: 'module' },
+              children: getPreambleCode(base),
+            },
+          ]
+      },
+      // In unbundled mode, Vite transforms any requests.
+      // But in full bundled mode, Vite only transforms / bundles the scripts injected in `order: 'pre'`.
+      order: isFullBundle ? 'pre' : undefined,
```
ArnaudBarre: Why does the preamble code to be "bundled"?

sapphi-red: It is because the preamble code has `import { injectIntoGlobalHook } from "/@react-refresh"`. `/@react-refresh` is not an entrypoint, so it's not possible to access without including the code in the bundle.

ArnaudBarre: Is there an issue to always run this 'pre'? Do we known knowledge of people depending on that?

sapphi-red: It will be a problem if you are using a plugin that transforms something else to an HTML (which should run in `order: 'pre'`).
I found this one from the code search: https://github.com/vituum/vite-plugin-pug/blob/ec22a2135fe05e9b0298c30a4fbf0b9ebe6e5803/index.js#L117-L122

ArnaudBarre (approved):

PR #962, opened 2025-11-04
fix(react-swc): fix `@vitejs/plugin-react-swc/preamble` on build

HcySunYang:
### Description

Should ensure hmrDisabled is set to true during build, preventing React refresh runtime from being included in production bundles.

Otherwise:

```✗ Build failed in 20ms
error during build:
[vite]: Rollup failed to resolve import "/@react-refresh" from "@vitejs/plugin-react-swc/preamble".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
```

### DIFF
--- a/packages/plugin-react-swc/src/index.ts
+++ b/packages/plugin-react-swc/src/index.ts
@@ -75,7 +75,7 @@ type Options = {
 }
 
 const react = (_options?: Options): Plugin[] => {
-  let hmrDisabled = false
+  let hmrDisabled = true
   let viteCacheRoot: string | undefined
   const options = {
     jsxImportSource: _options?.jsxImportSource ?? 'react',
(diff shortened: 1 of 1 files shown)


hi-ogawa (commented):
I think this is a bug fix? Probably I missed that `hmrDisabled = true` happens with `apply: 'serve'` currently https://github.com/vitejs/vite-plugin-react/blob/99d0fc443caf7c92c7f2771b29ecc784a1aa7240/packages/plugin-react-swc/src/index.ts#L119-L136

I think it should be safe to flip the logic by starting from `let hmrDisabled = true` and then check `config.server.hmr !== false`, which becomes closer to `@vitejs/plugin-react` https://github.com/vitejs/vite-plugin-react/blob/99d0fc443caf7c92c7f2771b29ecc784a1aa7240/packages/plugin-react/src/index.ts#L120

ArnaudBarre:
Yeah I would prefer to avoid an extra plugin which makes thinks harder to track

HcySunYang:
Have updated the implementation

ArnaudBarre (approved):

hi-ogawa (approved):
Thanks!

PR #1004, opened 2025-12-04
feat(rsc): support serialization of Request and Response with `loadModuleDevProxy`

jacob-ebey:
fix(rsc): allow error traces as it is server -> server communication

### Description

Allows for Request and Response classes to be sent through the RSC x-environment RPC mechanism to support more real-world setups such as react-router RSC in environments such as Cloudflare.


hi-ogawa (approved):
Thanks!

PR #1031, opened 2025-12-30
fix(rsc): use proper origin for source map lookup

hybrist:
### Description

When using RSCs inside of an `<iframe srcdoc="..."` environment, the origin is `null` and cannot be used as the 2nd argument to `new URL`. This causes an exception that completely breaks the page. This is far from a common setup but happens to be the environment for MCP Apps / ChatGPT Apps.

Using `import.meta.url` both fixes that immediate issue but also ensures that the URL properly respects `server.origin` if set in the configuration. Updated the existing TODO to focus on `BASE_URL` only.

### DIFF
--- a/packages/plugin-rsc/src/react/browser.ts
+++ b/packages/plugin-rsc/src/react/browser.ts
@@ -54,8 +54,11 @@ export function findSourceMapURL(
   filename: string,
   environmentName: string,
 ): string | null {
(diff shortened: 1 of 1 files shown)


hi-ogawa:
Thanks for the fix! Good to know window origin can disappear :+1:

hi-ogawa (approved):

PR #1044, opened 2026-01-10
fix(rsc): add ?direct to get actual CSS content in dev mode instead of JS module

schiller-manuel:


hi-ogawa:
I went deep dive into `?direct` mechanism in https://github.com/vitejs/vite-plugin-react/pull/1050 and I don't think this would matter for rsc plugin. As somewhat expected, css virtual module isn't working currently but there's no way around even if we incorporate `?direct` as far as I can tell (or we would need a bit more trickery or fix on Vite core side). I will explore a bit more, but will probably close this PR later.

hi-ogawa:
https://github.com/vitejs/vite-plugin-react/pull/1050 is merged, so closing.

PR #1054, opened 2026-01-15
test: lazy css reproduction

hi-ogawa:
Add reproduction for SSR-ing lazy client component which imports CSS. This is a test case for investigating FOUC (flash of unstyled content) when lazy-loaded client components import CSS files.

Ref: https://github.com/wakujs/waku/issues/1911#issuecomment-3752539573


hi-ogawa:
Superseded by #1056 (rebased on latest main)

PR #1062, opened 2026-01-15
chore(rsc): update starter-cf-single example

jamesopstad:
### Description

This updates the `starter-cf-single` example to use the new `childEnvironments` feature of `@cloudflare/vite-plugin` that was introduced in v1.21.0. This means that both the `rsc` and `ssr` environments now run in workerd in development.


hi-ogawa (approved):
Thanks!

PR #1065, opened 2026-01-16
feat(plugin-rsc): add callback to be notified when client reference dependencies are loaded

schiller-manuel:
## Summary

Add a callback mechanism to notify frameworks when client reference dependencies are loaded during SSR.

### New API

```ts
import { setOnClientReference, type OnClientReference } from '@vitejs/plugin-rsc/ssr'

const callback: OnClientReference = ({ id, deps }) => {
  // id: client reference module id
  // deps: { js: string[], css: string[] }
}

setOnClientReference(callback)
```

### Use case

During SSR, frameworks need to track which client components are actually rendered in order to collect their asset dependencies (JS and CSS) for the response. This callback is invoked every time a client reference is accessed, allowing per-request asset collection.

Note: The callback is called on every access (not memoized), so frameworks should handle deduplication if needed.

### Changes

- `setRequireModule` now accepts an optional `onLoad` callback
- New `setOnClientReference` function to register a framework-level callback
- New `OnClientReference` type export

All new APIs are marked `@experimental`.

### DIFF
--- a/packages/plugin-rsc/src/ssr.tsx
+++ b/packages/plugin-rsc/src/ssr.tsx
@@ -9,6 +9,29 @@ export { createServerConsumerManifest } from './core/ssr'
 
 export * from './react/ssr'
 
+/**
+ * Callback type for client reference dependency notifications.
+ * Called during SSR when a client component's dependencies are loaded.
+ */
+export type OnClientReferenceDeps = (deps: {
+  js: readonly string[]
+  css: readonly string[]
+}) => void
+
+// Registered callback for client reference deps
+let onClientReferenceDepsCallback: OnClientReferenceDeps | undefined
+
+/**
+ * Register a callback to be notified when client reference dependencies are loaded.
+ * Called during SSR when a client component is accessed.
+ *
+ */
+export function setOnClientReferenceDeps(
+  callback: OnClientReferenceDeps | undefined,
+): void {
+  onClientReferenceDepsCallback = callback
+}
+
 initialize()
 
 function initialize(): void {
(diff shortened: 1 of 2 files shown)


packages/plugin-rsc/src/ssr.tsx
```diff
@@ -38,6 +61,17 @@ function initialize(): void {
...
+    // Notify framework callback for per-request asset collection.
+    onLoad: (id) => {
+      if (!import.meta.env.__vite_rsc_build__) return
+      if (onClientReferenceDepsCallback) {
+        const deps = assetsManifest.clientReferenceDeps[id]
+        if (deps) {
+          onClientReferenceDepsCallback({ js: deps.js, css: deps.css })
+        }
+      }
+    },
```
hi-ogawa, line 75: Is this separate `onLoad` necessary? It looks like we can trigger the same thing on `// kick off preload` above inside `load`?

packages/plugin-rsc/src/ssr.tsx
```diff
@@ -9,6 +9,29 @@ export { createServerConsumerManifest } from './core/ssr'
...
+/**
+ * Register a callback to be notified when client reference dependencies are loaded.
+ * Called during SSR when a client component is accessed.
+ *
+ */
+export function setOnClientReferenceDeps(
+  callback: OnClientReferenceDeps | undefined,
+): void {
+  onClientReferenceDepsCallback = callback
+}
```
hi-ogawa, line 34: I think I'd expose fuller structure in the form of `onClientReference(id: string, deps: ...)`. Is it work for you?

hi-ogawa (commented):
The idea makes sense to me.
Actually the similar mechanism would be needed for Waku https://github.com/wakujs/waku/issues/1656. Waku's custom implementation before integrating `@vitejs/plugin-rsc` allowed this pattern, but I was hesitant to expose this since it will make it harder to upstream to React as it woudl be "non standard" rsc API. Again, I'm not sure what would be the long term effect of such API yet, so let me mark as `@experimental` and ship it.

hi-ogawa (approved):

PR #1070, opened 2026-01-19
fix(plugin-rsc): emit css deps in dev through onClientReference

schiller-manuel:


hi-ogawa:
I realized we can likely use the same mechanism as `wrapResourceProxy` for client reference notification. This is because module export (getter) access is not memoized by React (and also us) and that's the whole idea of `wrapResourceProxy` and `preloadDeps`. I'll probably create a new PR to remove `onLoad` mechanism in favor of that. Let me try.

https://github.com/vitejs/vite-plugin-react/blob/8dc7e5fa1191813bcfb9de1950a63ae289d10732/packages/plugin-rsc/src/ssr.tsx#L79-L80

hi-ogawa:
Can you test if this works for you? 
- https://github.com/vitejs/vite-plugin-react/pull/1073

PR #1074, opened 2026-01-20
fix(rsc): support `rolldownOptions`

hi-ogawa:
Add compatibility for Rolldown-based Vite by supporting both `build.rollupOptions.input` and `build.rolldownOptions.input`.

- Add `getBuildOptionsInput()` utility to read input from either option
- Add `createBuildInputConfig()` utility to generate config for both options
- Update `getEntrySource()` to use the new utility
- Update `ensureEnvironmentImportsEntryFallback()` to set both options
- Update environment config section to use `createBuildInputConfig()`
- Update `getFallbackRollupEntry()` usage to read from both options

This ensures the plugin works with both Rollup-based and Rolldown-based Vite versions as per the migration guide at main.vite.dev/guide/migration.

### DIFF
--- a/packages/plugin-rsc/src/plugin.ts
+++ b/packages/plugin-rsc/src/plugin.ts
@@ -45,7 +45,9 @@ import {
   parseReferenceValidationVirtual,
 } from './plugins/shared'
 import {
+  createBuildInputConfig,
   createVirtualPlugin,
+  getBuildOptionsInput,
   getEntrySource,
   hashString,
   normalizeRelativePath,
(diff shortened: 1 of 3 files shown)


hi-ogawa:
Actually supporting both `rolldownOptions` and `rollupOptions` from plugin side doesn't seem necessary since both already works thanks to Vite 8's own interop (i.e. Vite 7 supports only rollupOpitons and Vite 8 supports both rolldownOptions and rollupOptions). We just didn't have a test case for Vite 8 + rolldownOptions, so that's the only thing we need.

PR #1092, opened 2026-02-02
fix(plugin-rsc): cjs to esm interop helper doesn't handle native/external cjs import properly

### LINKED ISSUE #1072: RSC: Internal server error: (0 , html_dom_parser_1.default) is not a function
### Related plugins

### Describe the bug

Hello. I'm using Vite v8 in my project, and I'm gradually migrating to RSC, very exciting! I stumbled upon an interop bug with a transitive depenency of `html-react-parser` called `html-dom-parser`, and it happens only when I add the plugin, but only during development! It builds fine, so something is tripping up the module runner:

```
[vite] Internal server error: (0 , html_dom_parser_1.default) is not a function
      at HTMLReactParser (/Users/silvenon/Code/my-rsc-rewrite/node_modules/.pnpm/html-react-parser@5.2.11_@types+react@19.2.8_react@19.2.3/node_modules/html-react-parser/src/index.ts:35:14)
```

This is the only dependency I've had issues with so far related to the RSC plugin, but I also didn't get far in my migration success to try out all of my modules. Regarding my situation I managed to find a workaround — fortunately `html-react-parser` exports `html-dom-parser` primitives, so I was able to use those to achieve my goal.

The attached repro is not minimal, but it should be familiar to the maintainers because it's created from your `starter` example. The only thing I added is an additional call to the parse function in the SSR entry point to create the development error.

### Reproduction

https://github.com/silvenon/repro/tree/vite-rsc-ssr-no-external

### Steps to reproduce

```sh
# with GitHub CLI
gh repo clone silvenon/repro -b vite-rsc-ssr-no-external
cd repro
pnpm install
pnpm dev --open # error ❌
pnpm build
pnpm preview --open # works ✅
```

### System Info

```shell
System:
    OS: macOS 26.2
    CPU: (8) arm64 Apple M1 Pro
    Memory: 459.95 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.0 - /Users/silvenon/.n/bin/node
    npm: 10.9.4 - /Users/silvenon/.n/bin/npm
    pnpm: 10.28.0 - /Users/silvenon/.n/bin/pnpm
    bun: 1.1.29 - /Users/silvenon/.bun/bin/bun
    Deno: 1.45.2 - /opt/homebrew/bin/deno
  Browsers:
    Chrome: 144.0.7559.60
    Safari: 26.2
  npmPackages:
    @vitejs/plugin-react: ^5.1.2 => 5.1.2 
    @vitejs/plugin-rsc: ^0.5.14 => 0.5.14 
    vite: 8.0.0-beta.5 => 8.0.0-beta.8
```

### Used Package Manager

npm

### Logs

_No response_

### Validations

### LINKED ISSUE #1087: RSC: cjs to esm interop helper doesn't handle native/external cjs import properly
### Related plugins

## Description

The current `CJS_INTEROP_HELPER` in the CJS module runner transform doesn't properly handle built-in Node modules or external CJS modules.

## Current Behavior

The current implementation only checks if a module was transformed by this plugin:

\`\`\`javascript
function __cjs_interop__(m) {
  return m.__cjs_module_runner_transform ? m.default : m;
}
\`\`\`

## Problem

This is a minimal repro example

```javascript
const EventEmitter = require('events');

class CustomEventEmitter extends EventEmitter {
}
```

where the transformed esm result would be something like the following

```javascript
function __cjs_interop__(m) { return m.__cjs_module_runner_transform ? m.default : m; }

const EventEmitter = (__cjs_interop__(await import('events')));

class CustomEventEmitter extends EventEmitter {
}
```

when the transformed esm js file being evaluated, Node complains about `EventEmitter` because it's the namespace object not a constructor

```
Class extends value [object Module] is not a constructor or null
         at eval (...\node_modules\@repo\custom-event\index.js:3:27)
         at async ESModulesEvaluator.runInlinedModule (file:///.../node_modules/vite/dist/node/module-runner.js:913:3)
```

### Reproduction

https://github.com/ErleiX/vite-plugin-rsc-repro/

### Steps to reproduce

follow the repro steps in the readme file of the repro repo I link above

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 12th Gen Intel(R) Core(TM) i7-1265U
    Memory: 3.83 GB / 31.83 GB
  Binaries:
    Node: 24.13.0 - C:\nvm4w\nodejs\node.EXE
    npm: 11.6.2 - C:\nvm4w\nodejs\npm.CMD
  Browsers:
    Chrome: 144.0.7559.110
    Edge: Chromium (142.0.3595.94)
  npmPackages:
    @vitejs/plugin-react: ^5.1.2 => 5.1.2
    @vitejs/plugin-rsc: ^0.5.15 => 0.5.17
    vite: ^7.3.1 => 7.3.1
```

### Used Package Manager

npm

### Logs

Class extends value [object Module] is not a constructor or null
         at eval (...\node_modules\@repo\custom-event\index.js:3:27)
         at async ESModulesEvaluator.runInlinedModule (file:///.../node_modules/vite/dist/node/module-runner.js:913:3)

### Validations

ErleiX:
### Description

- Closes https://github.com/vitejs/vite-plugin-react/issues/1072
- Closes #1087

This change added a reproduce test for the problem that discovered by issue #1087

The test would fail without this fix and passes with this fix.

When it fails, it is exactly the same error described in issue #1087

The fix is to leverage the Node cjs to esm wrapping behavior, where the wrapped esm namespace of the cjs module is a duplicate of the default export, with additional properties of the default and its named export.

The fix is also considered "safe" and won't break the consumer code, because the fix is inheriently checking that the default is equivalent to the namespace object by comparing all the properties.


hi-ogawa:
Thanks for PR! Will take a look. Your test case looks good, but can you also explain what's the actual package you hit the issue?

packages/plugin-rsc/src/transforms/cjs.ts
```diff
@@ -9,9 +9,15 @@ import { analyze } from 'periscopic'
...
+// Unwrap .default for modules
+// 1. if it was transformed by this plugin (marked with __cjs_module_runner_transform)
+// 2. if all named exports point to .default properties (common CJS pattern)
+//    this is particularly important for Node built-in modules consumptions;
+//    where the built-in modules are not transformed by this plugin but still follow the CJS export pattern
+//    see [getESMFacade](https://github.com/nodejs/node/blob/f200685d9930404d610a52d9e06513bf0a821ed4/lib/internal/bootstrap/realm.js#L347-L360)
+//
 // This ensures we don't incorrectly unwrap .default on genuine ESM modules
-const CJS_INTEROP_HELPER = `function __cjs_interop__(m) { return m.__cjs_module_runner_transform ? m.default : m; }`
+const CJS_INTEROP_HELPER = `function __cjs_interop__(m) { return m?.__cjs_module_runner_transform ? m.default : m?.default && Object.keys(m).every(k => k === "default" || m[k] === m.default[k] || m[k] === m.default) ? m.default : m; }`
```
hi-ogawa: For `node:event` case, `m[k] === m.default` seems redundant. Do you have example where this matters?

```
if k is "default", k === "default" true

if k is not "default", then it's always 
  m[k] === m.default[k]
  which includes
  m.EventEmitter === m.default.EventEmitter 
```

so it won't reach the case of `m.EventEmitter === m.default`.

ErleiX: Good catch, we can remove the `m[k] === m.default`

ErleiX: I saw you've already did it, thanks!

ErleiX:
> Thanks for PR! Will take a look. Your test case looks good, but can you also explain what's the actual package you hit the issue?

sure thing! One of the real cases is that we have a dependency on the "ws" package, where the websocket/websocker-server.js is having exactly the same issue as the test case I added in the PR:

https://github.com/websockets/ws/blob/master/lib/websocket.js
https://github.com/websockets/ws/blob/master/lib/websocket-server.js

hi-ogawa:
Thanks for reference. In this case, `ws` package is importing `events`, but what I normally expect is to externalize `ws` since they don't need to go through `noExternal`. RSC plugin does aggresive `noExternal` only for react dependencies to ensure single `react` instance, but `ws` shouldn't be included there. Can you explain why `ws` ended up transformed by RSC's cjs plugin?

hi-ogawa:
That said, this fixes genuine bugs so I'm happy to merge this. Can you test pkg.pr.new in https://github.com/vitejs/vite-plugin-react/pull/1092#issuecomment-3838906208 to confirm if this fix is sufficient for your use case? (as I adjusted the interop slightly)

ErleiX:
> Thanks for reference. In this case, `ws` package is importing `events`, but what I normally expect is to externalize `ws` since they don't need to go through `noExternal`. RSC plugin does aggresive `noExternal` only for react dependencies to ensure single `react` instance, but `ws` shouldn't be included there. Can you explain why `ws` ended up transformed by RSC's cjs plugin?

One of the things the team decided is to bundling everything for ssr/rsc to make the server side deployment easier(and potentially better performance without having to resolve the external dependencies at runtime), that's why our vite config is very aggressive with "noExternal". I am happy to learn your perspective on what the benefits or trade-offs are by making dependencies like `ws` external

hi-ogawa:
Right, bundling server output is actually legitimate as far as I know, but that can be done at separate build pipeline after vite build. Alternatively, you can also configure `noExternal` only for `vite build`, but not on `vite dev` since inherently transforming all server dependencies can be slower than let node import it as external. This is possible by conditionally returning config via `defineConfig(env => env.command === 'build' ? ... : ...)` https://vite.dev/config/#conditional-config.

ErleiX:
> Right, bundling server output is actually legitimate as far as I know, but that can be done at separate build pipeline after vite build. Alternatively, you can also configure `noExternal` only for `vite build`, but not on `vite dev` since inherently transforming all server dependencies can be slower than let node import it as external. This is possible by conditionally returning config via `defineConfig(env => env.command === 'build' ? ... : ...)` https://vite.dev/config/#conditional-config.

that seems to be a better solution, let me try that out, thanks for the suggestion. I also realized that Vite dev and build modes are quite different during debugging of this issue, which makes me wonder what are other gotchas that we'd better to be aware and config dev mode intentionally different, if there are such, please kindly share, thanks!

hi-ogawa (approved):

ErleiX:
> That said, this fixes genuine bugs so I'm happy to merge this. Can you test pkg.pr.new in [#1092 (comment)](https://github.com/vitejs/vite-plugin-react/pull/1092#issuecomment-3838906208) to confirm if this fix is sufficient for your use case? (as I adjusted the interop slightly)

confirming that this fix is sufficient for my use case, please feel free to merge it

hi-ogawa:
Thanks for contribution!

hi-ogawa:
The fix is released in v0.5.18 https://github.com/vitejs/vite-plugin-react/releases/tag/plugin-rsc%400.5.18

PR #1096, opened 2026-02-05
fix(plugin-rsc): avoid stripping rolldown runtime during scan build

sapphi-red:
### Description

fixes this ecosystem-ci failure: https://github.com/vitejs/vite-ecosystem-ci/actions/runs/21699933912/job/62578172448#step:7:952

refs https://github.com/rolldown/rolldown/issues/8202


hi-ogawa (approved):
Looks good.

One question, this probably means that the existence of `\0rolldown/runtime.js` (also without `\0` in resolve?) and it's coming through user land plugin is a sort of public API? Is the documentation available now?

sapphi-red:
Good point; There's no document now. I've made an issue: https://github.com/rolldown/rolldown/issues/8204

PR #1106, opened 2026-02-13
docs(react-swc): deprecate plugin react swc

sapphi-red:
### Description

I wrote a migration guide and tried it on the playgrounds.

refs https://github.com/vitejs/vite/pull/21633

### DIFF
--- a/packages/plugin-react-swc/README.md
+++ b/packages/plugin-react-swc/README.md
@@ -1,5 +1,104 @@
 # @vitejs/plugin-react-swc [![npm](https://img.shields.io/npm/v/@vitejs/plugin-react-swc)](https://www.npmjs.com/package/@vitejs/plugin-react-swc)
 
+> ![CAUTION]
+> This plugin is now deprecated.
+>
+> Vite 8 now uses [Oxc](https://oxc.rs/) for fast refresh transform.
+> If you do not need other SWC-specific transforms, you can use [`@vitejs/plugin-react`](https://github.com/vitejs/vite/tree/main/packages/plugin-react) instead.
+>
+> If you were using this plugin without any options or only the following options:
+>
+> - `jsxImportSource`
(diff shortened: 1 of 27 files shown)


packages/plugin-react-swc/README.md
```diff
@@ -1,5 +1,104 @@
...
+>                 jsx: ext === 'tsx' || ext === 'jsx',
+>               },
+>               experimental: {
+>                 // if `plugins` were set
+>                 plugins: [['@swc/plugin-emotion', {}]],
+>               },
+>               transform: {
+>                 useDefineForClassFields: true,
+>                 react: {
+>                   runtime: 'preserve',
```
ArnaudBarre, line 83: you do need `importSource: 'reactlike'` here

ArnaudBarre (commented):
I'm ok with having a note that invite users to use OXC when possible, but as soon as you need SWC transform, I think it adds a lot of complexity for the users to handle this.
Plus, even if it will probably be imperceptible (if source map merging is fast), I think running both SWC + OXC is not ideal and can create confusions (it's not obvious that jsxRuntime should be passed to SWC and not OXC).

packages/plugin-react-swc/README.md
sapphi-red: The idea was to leave things as-is as much as possible in the SWC pass (TS -> JS is done because SWC doesn't have an option to disable it). We can do it in the opposite way as you suggested.

sapphi-red:
We can totally keep this plugin. When I started writing, I didn't expect the config to be this much complex.

packages/plugin-react-swc/README.md
ArnaudBarre: Oh yeah I missed the 'preserve'
Yeah it makes it a bit more confusing to have TSX->JSX in SWC and then JSX->JS with oxc

PR #1112, opened 2026-02-16
fix(rsc): copy only imported css and assets from rsc environemnt to client environment

### LINKED ISSUE #1109: Rework the mechanism of copying server asset as client asset copy
### Related plugins

### Description

Currently rsc plugin copies all assets from `rsc` environment to `client` environment by default.

https://github.com/vitejs/vite-plugin-react/blob/f066114c3e6bf18f5209ff3d3ef6bf1ab46d3866/packages/plugin-rsc/src/plugin.ts#L1078-L1087 

This is mainly for copying css assets on server that are handled internally by rsc plugin but also end-user's `?url` import inside `rsc` environment.

However this is not a safe heuristics since `rsc` environment is the server logic which should get exposed as public assets. As one such example, `rscViteManifest` is avoided to get copied above. 

For this concern, rsc plugin provides `copyServerAssetsToClient` option to control which gets copied, but this option isn't composable and more like targetted as end-user's tweak to filter out something.

https://github.com/vitejs/vite-plugin-react/blob/f066114c3e6bf18f5209ff3d3ef6bf1ab46d3866/packages/plugin-rsc/src/plugin.ts#L190-L195

### Suggested solution

Explore whether we can improve default heuristics to only detect assets which are clearly intended as public assets i.e. side effect css import and `?url` import in rsc environment.

If user plugins employs `emitFile({ type: "asset" })` in rsc environment, they shouldn't get copied to public assets in client environment by default, but that should be opt-in behavior. For example, Astro seems to have introduced `emitClientAsset` https://github.com/withastro/astro/pull/15345 for such behavior possible.

### Alternative

_No response_

### Additional context

_No response_

### Validations

hi-ogawa:
### Description

- Closes https://github.com/vitejs/vite-plugin-react/issues/1109

The detail documented in `packages/plugin-rsc/docs/notes/2026-02-16-issue-1109-server-asset-copy-plan.md`.

It turns out `chunk.viteMetadta.importedCss/importedAssets` is exactly what we need. We have a test case to cover cases like:
- css side effect import 
- `?url` css
- nested assets through `.css` and `url()`
- `rsc` environment `emitFile` is not included in public asset 

So, if tests are passing, this should be the right behavior. We might consider opt-in public `emitFile` in server environment, but that seems unnecessary for the time being.

### DIFF
--- a/packages/plugin-rsc/docs/notes/2026-02-16-issue-1109-server-asset-copy-plan.md
+++ b/packages/plugin-rsc/docs/notes/2026-02-16-issue-1109-server-asset-copy-plan.md
@@ -10,14 +10,16 @@ That behavior is convenient, but unsafe as a default:
 - it relies on a coarse heuristic (`copyServerAssetsToClient ?? (() => true)`)
 - it already needs special-case exclusions (`.vite/manifest.json`), which is a smell
 
-## Implementation status
+## Implementation status (final)
 
-Implemented on 2026-02-16.
+Implemented on 2026-02-16 with follow-up refinements.
 
-- Added `collectPublicServerAssets(bundle)` in `packages/plugin-rsc/src/plugin.ts`.
-- Updated `generateBundle` copy logic to use that set as the default filter.
-- `copyServerAssetsToClient` is now deprecated and treated as a no-op.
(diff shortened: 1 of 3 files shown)


hi-ogawa:
@jamesopstad Can you test if this works for you? :pray:

jamesopstad:
> @jamesopstad Can you test if this works for you? 🙏

I've tested and this works. Thanks!

PR #1113, opened 2026-02-17
chore(rsc): update starter-cf-single to remove `serverHandler: false`

jamesopstad:
### Description

Removes `serverHandler: false` from the Cloudflare example now that this is set internally by `@cloudflare/vite-plugin`.


hi-ogawa (approved):
Thank you!

PR #1123, opened 2026-02-27
feat(react)!: remove babel

sapphi-red:
### Description

~~_**Built on top of #1124**_~~

In Vite 8 (currently beta), react refresh transform can be handled by Oxc. This means there's no transform that uses babel by default.
To reduce the install size, I think we should remove babel from the dependency of `@vitejs/plugin-react`.

The `@rolldown/plugin-babel` package lives at https://github.com/rolldown/plugins.

context: https://github.com/vitejs/vite-plugin-react/discussions/1114


playground/include-node-modules/vite.config.ts
```diff
@@ -1,25 +1,26 @@
-import type { PluginItem as BabelPlugin } from '@babel/core'
+import babel from '@rolldown/plugin-babel'
 import react from '@vitejs/plugin-react'
 import { defineConfig } from 'vite'
 
 export default defineConfig({
   plugins: [
     react({
       exclude: [/\/node_modules\/(?!(\.pnpm\/)?test-package)/],
```
ArnaudBarre, line 12: I don't think this is need given the current test.
We are now testing the capability of `@rolldown/plugin-babel` to include node_modules.
Maybe the package could contain JSX to show that we can force OXC to run on node_modules files?

packages/plugin-react/src/index.ts
```diff
@@ -444,81 +244,14 @@ export default function viteReact(opts: Options = {}): Plugin[] {
 
 viteReact.preambleCode = preambleCode
 
+export { reactCompilerPreset }
+
 // Compat for require
 function viteReactForCjs(this: unknown, options: Options): Plugin[] {
   return viteReact.call(this, options)
 }
 Object.assign(viteReactForCjs, {
```
ArnaudBarre, line 253: If this needed when supporting only Vite 8+ which means require(esm) IIRM?

playground/compiler/package.json
```diff
@@ -12,7 +12,9 @@
     "react-dom": "^19.2.4"
   },
   "devDependencies": {
+    "@babel/core": "^7.29.0",
     "@babel/plugin-transform-react-jsx-development": "^7.27.1",
```
ArnaudBarre: Is this plugin needed? It's not here in multiple playground
Is it injected by `@rolldown/plugin-babel` ?

ArnaudBarre (commented):
Looks cool, thanks for the cleanup, I'm happy I will be able stop downloading babel!
We need to updates the docs (both Readmes in the repo and vite.dev)
I think we should also add a section on how to add babel to a Vite 8 project, promoting this new plugin and all the possibilities for filtering.

packages/plugin-rsc/e2e/react-compiler.test.ts
```diff
@@ -1,28 +1,39 @@
...
+  // use RUNNER_TEMP on Github Actions
+  // https://github.com/actions/toolkit/issues/518
+  const tmpRoot = path.join(
+    process.env['RUNNER_TEMP'] || os.tmpdir(),
+    'test-vite-rsc-react-compiler',
+  )
 
   test.beforeAll(async () => {
-    await setupInlineFixture({
+    await setupIsolatedFixture({
```
hi-ogawa: I think this pattern should work fine with `setupInlineFixture`. What was broken?
We probably just need to add `@rolldown/plugin-babel` to `packages/plugin-rsc/examples/e2e/package.json`.

sapphi-red: Ah, I overlooked that it runs in `examples/e2e/temp`. I wrongly assumed that it runs on `examples/starter`.

playground/compiler/package.json
sapphi-red: Ah, it can be removed.

playground/include-node-modules/vite.config.ts
sapphi-red: Good catch. I've updated the test to ensure that the react refresh transforms are applied and the HMR works.

packages/plugin-react/src/index.ts
sapphi-red: This is needed to keep the value of `require('@vitejs/plugin-react')`. If we remove this, the config has to be:
```js
const react = require('@vitejs/plugin-react')

module.exports = {
  plugins: [react.default()] // instead of `react()`
}
```

packages/plugin-rsc/src/plugin.ts
```diff
@@ -1352,7 +1352,7 @@ function vitePluginUseClient(
     // path in metafile is relative to cwd
     // https://github.com/vitejs/vite/blob/dd96c2cd831ecba3874458b318ad4f0a7f173736/packages/vite/src/node/optimizer/index.ts#L644
     id = normalizePath(path.relative(process.cwd(), id))
-    if (optimizerMetadata?.ids.includes(id)) {
+    if (optimizerMetadata?.ids.map((id) => normalizePath(id)).includes(id)) {
```
hi-ogawa: I moved the fix in `optimizerPluginRolldown` itself in the last PR. Looks like a stale diff.

packages/plugin-react/README.md
```diff
@@ -68,56 +67,44 @@ By default, the plugin uses the [automatic JSX runtime](https://legacy.reactjs.o
...
 ```
 
-Note: When not using plugins, only esbuild is used for production builds, resulting in faster builds.
+Under the hood, this simply updates the React Fash Refresh runtime URL from `/@react-refresh` to `http://localhost:3000/@react-refresh` to ensure there is only one Refresh runtime across the whole application. Note that if you define `base` option in the host application, you need to include it in the option, like: `http://localhost:3000/{base}`.
 
-#### Proposed syntax
+## React Compiler
 
-If you are using ES syntax that are still in proposal status (e.g. class properties), you can selectively enable them with the `babel.parserOpts.plugins` option:
+React Compiler support is available via the exported `reactCompilerPreset` helper, which requires `@rolldown/plugin-babel` and `babel-plugin-react-compiler` as peer dependencies:
```
hi-ogawa: Can we add links to other project? I'd expect "React Compiler", "@rolldown/plugin-babel", "babel-plugin-react-compiler" to be links to repo or their docs.

packages/plugin-react/README.md
```diff
@@ -68,56 +67,44 @@ By default, the plugin uses the [automatic JSX runtime](https://legacy.reactjs.o
...
-**Note:** TypeScript syntax is handled automatically.
+```js
+// vite.config.js
+import { defineConfig } from 'vite'
+import react, { reactCompilerPreset } from '@vitejs/plugin-react'
+import babel from '@rolldown/plugin-babel'
 
-Here's the [complete list of Babel parser plugins](https://babeljs.io/docs/en/babel-parser#ecmascript-proposalshttpsgithubcombabelproposals).
+export default defineConfig({
+  plugins: [react(), babel(reactCompilerPreset())],
```
hi-ogawa: Isn't this `babel({ presets: [reactCompilerPreset()] })` or is this really supposed?

sapphi-red: My bad, it slipped through my eyes. Fixed 👍

packages/plugin-react/tsconfig.json
```diff
@@ -10,6 +10,8 @@
     "sourceMap": true,
     "noEmit": true,
     "noUnusedLocals": true,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "rootDir": ".",
```
hi-ogawa, line 14: Is `rootDir` needed? I thought this would allow importing it like `import "src/any-file-inside-project"` without type error and I didn't like that before.

sapphi-red: It is needed, otherwise I get the following error:

> error TS2210: The project root is ambiguous, but is required to resolve import map entry '.' in file 'path/to/vite-plugin-react/packages/plugin-react/package.json'. Supply the `rootDir` compiler option to disambiguate.

> I thought this would allow importing it like `import "src/any-file-inside-project"` without type error

I guess you are confused with [`baseUrl`](https://www.typescriptlang.org/tsconfig/#baseUrl).

hi-ogawa: Ah good to know. Indeed, `baseUrl` is the one I hated :sweat_smile:

hi-ogawa (approved):

sapphi-red:
I'll cut a beta version with this PR.

PR #1124, opened 2026-02-27
refactor(react)!: drop Vite 7 and below support

sapphi-red:
### Description

Drops Vite 7 and below support from plugin-react.

As a preparation for #1123.


packages/plugin-rsc/src/plugin.ts
```diff
@@ -1352,7 +1352,7 @@ function vitePluginUseClient(
     // path in metafile is relative to cwd
     // https://github.com/vitejs/vite/blob/dd96c2cd831ecba3874458b318ad4f0a7f173736/packages/vite/src/node/optimizer/index.ts#L644
     id = normalizePath(path.relative(process.cwd(), id))
-    if (optimizerMetadata?.ids.includes(id)) {
+    if (optimizerMetadata?.ids.map((id) => normalizePath(id)).includes(id)) {
```
sapphi-red: This change is needed probably because of the change in https://github.com/vitejs/vite/pull/21720

ArnaudBarre (approved):

hi-ogawa (commented):
I might want a way to keep testing Vite 7 on CI. Let me think a bit.

packages/plugin-rsc/src/plugin.ts
hi-ogawa: To be honest, the original code looks bad. We should've had `normalizePath` right after `path.relative` in `optimizerPluginRolldown` below. I moved the fix to there.

hi-ogawa (approved):
I added Vite 7 run in rsc ci.

PR #1132, opened 2026-03-02
chore(deps): test tailwindcss patch

hi-ogawa:
### Description

Testing the build from https://github.com/tailwindlabs/tailwindcss/pull/19745.


packages/plugin-rsc/e2e/basic.test.ts
hi-ogawa, line 1: I don't remember much about the context of this test, but it looks like tailwind forces reloading on `unuded.tsx` changes seems reasonable. This looks like independent from SSR framework issue of tailwind, so let's adjust the test case here:
- https://github.com/vitejs/vite-plugin-react/pull/658

hi-ogawa:
The fix is merged upstream. Closing

PR #1137, opened 2026-03-05
docs(react): clarify react compiler preset filter

### LINKED ISSUE #1135: Possibly greedy reactCompilerPreset regex
I don't fully understand what this regex is trying to match?

The `\b[A-Z]` part matches almost any capital letter. `\buse` matches for example `'use-credentials'`.

Is it meant to match React components and hooks? If so it's a bit greedy. Anyway, thanks for maintaining this package!

Regex101 link: https://regex101.com/r/VTYx5E/1

https://github.com/vitejs/vite-plugin-react/blob/704e0d388b3ecf56d08d3c7aa128c832d38f497e/packages/plugin-react/src/reactCompilerPreset.ts#L20

sapphi-red:
close #1135

### DIFF
--- a/packages/plugin-react/README.md
+++ b/packages/plugin-react/README.md
@@ -107,6 +107,19 @@ babel({
 })
 ```
 
+> [!TIP]
+>
+> `reactCompilerPreset` is only a convenient helper with a preconfigured filter. You can configure override the filters to fit your project structure or code. For example, if you know a large portion of your files are never React/hook-related or won't benefit from the React Compiler, you can aggressively exclude them via `rolldown.filter`:
+>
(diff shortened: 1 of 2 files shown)


hi-ogawa (approved):

hi-ogawa:
I was also thinking to directly configure `@rolldown/plugin-babel` with `babel-plugin-react-compiler` without touching presets.  That should be also possible right?

sapphi-red:
That's also possible: https://github.com/rolldown/plugins/tree/main/packages/babel#rolldown-babel-presets

ArnaudBarre (approved):

PR #1138, opened 2026-03-05
perf(react): improve react compiler preset so that slightly more modules are filtered out

sapphi-red:
I noticed that we can add `[A-Z0-9]` to filter out a bit more modules.

https://github.com/facebook/react/blob/9c0323e2cf9be543d6eaa44419598af56922603f/compiler/packages/babel-plugin-react-compiler/src/Entrypoint/Program.ts#L897-L899


hi-ogawa (approved):

packages/plugin-react/src/reactCompilerPreset.ts
```diff
@@ -17,7 +17,7 @@ export const reactCompilerPreset = (
       code:
         options.compilationMode === 'annotation'
           ? /['"]use memo['"]/
-          : /\b[A-Z]|\buse/,
+          : /\b[A-Z]|\buse[A-Z0-9]/,
```
ArnaudBarre: I think the only way to define a function without a whitespace before is `const A,B= ...` which I don't know why you would do that in anyworld for defining hooks or components
```suggestion
          : /\s[A-Z]|\suse[A-Z0-9]/,
```

But I think the better tradeoff is always running on `jsx|tsx|mdx` (in a normal codebase, you have like 98/99% of jsx files that contrains components, so not sure trying to get a 2% improvement with a regex is worth it) and run on `js|ts` only with when code matches `\suse[A-Z0-9]` (Vite always ban using JSX inside JS, nobody writes manual `jsx` calls and plugin should not rename file ids once transform from jsx to js

The only issue with the approach if for people having custom DSL that compile to React, but I think these frameworks should provide their own react compiler preset 

Also can you put the link from the PR into the source code? I think it's good to have it as a reference if someone challenges it later

sapphi-red: If it's possible to have `,` before the name, I think we can use `[\s,]`. I guess `\s` and `[\s,]` won't have much perf difference.

> I think the better tradeoff is always running on `jsx|tsx|mdx` (in a normal codebase, you have like 98/99% of jsx files that contrains components, so not sure trying to get a 2% improvement with a regex is worth it) and run on `js|ts` only with when code matches `\suse[A-Z0-9]` (Vite always ban using JSX inside JS, nobody writes manual `jsx` calls and plugin should not rename file ids once transform from jsx to js

It's currently difficult to achieve, but it would be easier to do once Vite implements [composable filters](https://rolldown.rs/apis/plugin-api/hook-filters#composable-filters) (https://github.com/vitejs/rolldown-vite/issues/605).

ArnaudBarre: After some AI checking, I think this is pretty safe and code not matching this regex will certainly break a lot a React rules:
`(?:const|let|var|function)\s+(?:[A-Z]|use[A-Z0-9])`

(The way to break it is using `new Function` or putting your component in a class method, both sounds horrible)

Speaking of classes, I checked and the playground doesn't touch class component, so no need to match this

Edit: Ok no actually this would be valid IMO but the [playground](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwBUYCAyngIZ4IEC+BAZjBBgQDogwJVx5cA3BwB2ouBBFhCACwgQA1mAIBeIqIKbSCACoJpyAgAoAlKoB86kZs0SphANrSaCADQEyeSi4C6qkmTetEYADCbC1jZoTMbOtKoqamEEPHiw1gA8ACZoAG7mAHIQeDJoIgDmGQD0OfkaNggANmQpCGkwmbXmwHEI9NVd9fSu9dp60gBMhkxQIvxokgG6+ngTplY2BHbSBE7UtO6eQQh+atrHoeH1mtGx+3SJSWap6QTZeYXFpRUDH9cETRaLw6by6PXu-Rqf0iw1E9FEGm2hDGK386xUlmA1yRu16hza3hO-nO90uESiMSMvQST1a7U6HyKJTKlShdUigLowIZ+XBLkhgxE9BArhAdiYaHKKHQ2FwhDwAE8sHRiAAFRpQcplADyWDwCykDGYrHYAHIAEZUc1NAC0WA1WpENp4fDwNok2DQjQQMCh0lNEVERixkSqVQ9WC9NANAFkIFkEIYuFRGo0uHCPNGwBL9AR1ZqdXqDWBwiLwHIAO4ASREtA6KbAKCYDb6QA) doesn't touch it:

```jsx
import { useState } from "react";

const hooks = {
  useTest: () => {
    const [state, setState] = useState(0);
    if (state === 0) return <div>Nothing</div>
    else return <div>{state}</div>
  }
}
```

I think the React compiler is only looking for top-level functions, and this sounds reasonable, so we can make an more aggressive regex

sapphi-red: > The way to break it is using `new Function` or putting your component in a class method, both sounds horrible

I guess `new Function` cannot be handled by the compiler anyways. Classes (& class methods) are skipped [here](https://github.com/facebook/react/blob/9c0323e2cf9be543d6eaa44419598af56922603f/compiler/packages/babel-plugin-react-compiler/src/Entrypoint/Program.ts#L530-L544).

> I think the React compiler is only looking for top-level functions

In `compilationMode: 'all'`, it seems to [only look for top-level ones](https://github.com/facebook/react/blob/9c0323e2cf9be543d6eaa44419598af56922603f/compiler/packages/babel-plugin-react-compiler/src/Entrypoint/Program.ts#L502-L508). But I guess it does otherwise.

Something like this won't match the regex you suggested, while the compiler optimizes:
```jsx
import React, { useState } from "react";
import { jsx } from "react/jsx-runtime"

export const components = {
  A: React.memo(() => {
    const [state, setStae] = useState(0);

    return jsx("div", { children: state })
  })
}
```
[playground](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEASggIZx4A0BwBUYCAynqXggQL4EBmMEGBADogYZCsIDcggHaYc+GgQBWYAB6cefAcNHk8AehWqAtDCjS8mBMJkyEq+YTgRpYJ-xzSEFsAQC8NDIEBACCyMRieAB0GAgYEAAUCQCU-gB8gdLBwc6uhADabqwI1Ax4zKQIALr+dAwVbAkADMlS0kHZoniwWUYJwgAmaABuwtS0cAAWaAA2A6LS4UVsnMkdHGvSHDIglCC53GgA5ijo2LiEeACeWOy0AAozUEdo0gDyWJYuvly8-AQAcgARqQgQgZsYsE8XtJTJFjM5sLMEDB9EM3AC2jIEsAOvp9IisLNWGgXABZCADBDhYSkGYzGxbAhgElgQ4IXyPZ6vD5fVytXbgSYQADuAEkLCjpHSwChuDKEBwgA)
This case matches [this part](https://github.com/facebook/react/blob/9c0323e2cf9be543d6eaa44419598af56922603f/compiler/packages/babel-plugin-react-compiler/src/Entrypoint/Program.ts#L1067-L1076).

That said, I think we can ignore this case if the speed up is big enough.

ArnaudBarre: It feels quite unexpected that the component is optimized only if wrap a in a specific function.

We could also add `["']react['"]` to the regex in that case, it should cover a lot of extra cases like this while still skipping files that are unrelated to React

ArnaudBarre: @sapphi-red so what about `["']react['"]|(?:const|let|var|function)\s+(?:[A-Z]|use[A-Z0-9])`?

sapphi-red: I added tests to check if the regex covers all the cases and found that it's possible to alias `memo` / `forwardRef`, so I replaced `["']react['"]` with `forwardRef|memo`. I also changed the latter part to match with `const components = { MyComponent() { return <></> } }`.

So in the end, it became `/forwardRef|memo|function\s+(?:[A-Z]|use[A-Z0-9])|(?:[A-Z]|use[A-Z0-9])[^\s:=(){}[\],;]*\s*[:=(]/`. WDYT?

ArnaudBarre: The issue is that it matches any files containing an object with an uppercase key.
We can keep it on the safe side until we are able to force run on every JSX files and we will be able to be a lot stricter on what matches in non JSX files

sapphi-red: Tweaked a bit to exclude those cases.

> We can keep it on the safe side

Do you mean to keep the implementation in main branch?

ArnaudBarre: No I would call this one safe because it's really try to match cases where functions are anonymous, which should be avoided and linted against

ArnaudBarre: Because top-level and nested are matched with the same part of the regex, now it doesn't work for top level component with complex type annotation:

```tsx
const X: SomeComplexType<Generic, number> = () => {
  return <></>
}
```

sapphi-red: Ah, I forgot that we need to handle TS code. I've fixed that now.
I also noticed that this filter doesn't work when there're comments in between.

ArnaudBarre: I think you actively trying to break your tools if you use comments between `const`/`function` and the function name so I would not bother handling these

sapphi-red: yeah, I think the same

ArnaudBarre (approved):

PR #1141, opened 2026-03-12
docs: add a link to the Oxlint rule for component exports alongside the ESLint rule link.

connorshea:
### Description

The rule from `eslint-plugin-react-refresh` - for only exporting components - is also available as a ported Rust rule in Oxlint, and I figure it makes sense to link it in the docs for users of Oxlint.

Also fixes the capitalization for ESLint :)


ArnaudBarre (approved):

ArnaudBarre:
BTW I did a bit rewrite of the rule in `0.5` to catch more issues and have less false positive. I don't know if this kind of think is tracked on oxlint side. One day I will take time to finish migrating from ESLint of oxlint, so I might do it myself one day 😄

PR #1146, opened 2026-03-13
feat(plugin-react): expand `@rolldown/plugin-babel` peer dep range

sapphi-red:


xoticmind (commented):
[plugin:vite:react-babel]

PR #1169, opened 2026-03-31
fix(rsc): align export async function checks

james-elicx:
### Description

The wrap-export and proxy-export files were both checking async functions slightly differently. The proxy-export file's version was less permissive than the wrap-export one. This changes them to share the same logic (and adds class expressions as well at the same time).

I believe this was the cause of https://github.com/cloudflare/vinext/issues/722 and should potentially fix it.


hi-ogawa (approved):
Thanks for the fix! I move the util and unify the two test cases.

PR #1183, opened 2026-04-12
fix(rsc): preserve `*` for generators

james-elicx:
The `*` syntax for generator functions was being dropped, resulting in syntax errors. This preserves it.


hi-ogawa (approved):
Looks good. Thanks!

PR #1189, opened 2026-04-16
feat(react): whitelist debugging-options

annitya:
### Description

tldr: Options should allow debug-settings. 

Changes in code can easily cause a component to be skipped over.
You might not notice, but infinite loops or self-ddos can also happen. 

Options added from [debugging-docs](https://react.dev/reference/react-compiler/configuration#debugging).

### DIFF
--- a/packages/plugin-react/README.md
+++ b/packages/plugin-react/README.md
@@ -107,6 +107,8 @@ The `reactCompilerPreset` accepts an optional options object with the following 
 - `compilationMode` — Set to `'annotation'` to only compile components annotated with `"use memo"`.
 - `target` — Set to `'17'` or `'18'` to target older React versions (uses `react-compiler-runtime` instead of `react/compiler-runtime`).
 
+Additional options can be found in the [documentation](https://react.dev/reference/react-compiler/configuration).
+
 ```js
 babel({
   presets: [reactCompilerPreset({ compilationMode: 'annotation' })],
--- a/packages/plugin-react/src/reactCompilerPreset.ts
+++ b/packages/plugin-react/src/reactCompilerPreset.ts
@@ -4,10 +4,7 @@ import type {
 } from '#optionalTypes'
 
 export const reactCompilerPreset = (
-  options: Pick<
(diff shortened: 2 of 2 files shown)


packages/plugin-react/src/reactCompilerPreset.ts
```diff
@@ -6,7 +6,7 @@ import type {
 export const reactCompilerPreset = (
   options: Pick<
```
ArnaudBarre: @sapphi-red Do we need `Pick`? I don't see a reason to allow all new options in the future

packages/plugin-react/README.md
```diff
@@ -106,6 +106,8 @@ The `reactCompilerPreset` accepts an optional options object with the following
 
 - `compilationMode` — Set to `'annotation'` to only compile components annotated with `"use memo"`.
 - `target` — Set to `'17'` or `'18'` to target older React versions (uses `react-compiler-runtime` instead of `react/compiler-runtime`).
+- `logger` — Example: `{ logEvent(filename, event) { console.log(`[Compiler] ${event.kind}: ${filename}`); } }`
```
ArnaudBarre: Maybe we can link to [this page](https://react.dev/reference/react-compiler/configuration) after those two main ones so we don't need to sync in the future

packages/plugin-react/src/reactCompilerPreset.ts
sapphi-red: I added `Pick` because I thought the other options were internal ones that were not expected to be used because I didn't find the documentations for them. But it turns out there are documents.

annitya: I assumed some of the options either had side-effects or were untested. Should I just remove the Pick and update the docs?

ArnaudBarre: Yes you can remove the Pick and update the readme to link to the official docs for 'more options' (compilationMode & target)

ArnaudBarre (approved):

PR #1192, opened 2026-04-16
fix(rsc): include bundled server CSS when cssCodeSplit is false

schiller-manuel:


hi-ogawa (approved):
Thanks for the fix 👍
I simplified a bit with mostly the same idea.

PR #1197, opened 2026-04-22
fix(plugin-react): use '/' base in bundledDev preamble to fix non-root base paths

### LINKED ISSUE #1190: bundledDev: getPreambleCode uses config.base, breaking non-root base paths
## Description

When using `experimental: { bundledDev: true }` with a non-root `base` (e.g. `base: "/static/"`), the `vite:react-refresh-fbm` plugin injects a preamble with `getPreambleCode(base)`, producing:

```js
import { injectIntoGlobalHook } from "/static/@react-refresh";
```

Rolldown tries to resolve `/static/@react-refresh` as a module specifier and fails, since the `resolveId` hook only matches the exact string `/@react-refresh`.

In non-bundled dev mode this works because the browser fetches it as a URL and Vite's `baseMiddleware` strips the prefix before the request reaches the resolve/load hooks. But in bundled dev mode, Rolldown resolves module specifiers at build time — there's no URL-level base stripping.

## Reproduction

```ts
// vite.config.ts
export default defineConfig({
  base: "/static/",
  experimental: { bundledDev: true },
  // ...
})
```

```
[UNRESOLVED_IMPORT] Error: Could not resolve '/static/@react-refresh' in entries/app/index.html?html-proxy&index=0.js
```

## Expected Behavior

The bundled dev mode preamble should use `getPreambleCode("/")` (or just `"/@react-refresh"` directly), since Rolldown resolves module specifiers, not URLs.

## Environment

- `vite`: 8.0.8
- `@vitejs/plugin-react`: 6.0.0

JustInCache:
## Summary

Fixes #1190

When using `experimental: { bundledDev: true }` with a non-root `base` (e.g. `base: "/static/"`), the `vite:react-refresh-fbm` plugin was calling `getPreambleCode(base)`, producing:

```js
import { injectIntoGlobalHook } from "/static/@react-refresh";
```

Rolldown resolves module specifiers at **build time** — there is no URL-level base-stripping middleware involved. So `/static/@react-refresh` never matched the `resolveId` hook which only handles the exact string `/@react-refresh`, causing:

```
[UNRESOLVED_IMPORT] Error: Could not resolve '/static/@react-refresh'
```

In non-bundled dev mode this worked because the browser fetched the URL and Vite's `baseMiddleware` stripped the prefix before the request reached the resolve/load hooks.

## Fix

Pass `'/'` instead of `config.base` to `getPreambleCode` inside `viteReactRefreshBundledDevMode`. This produces `/@react-refresh` directly, which the `resolveId` hook can match. This is consistent with what `virtualPreamblePlugin` already does (with the comment *"vite dev import analysis can rewrite base"*).

## Test

Added a new e2e playground `playground/bundled-dev-base-path` with `base: '/static/'` + `experimental: { bundledDev: true }` that verifies:
- The page loads without any `UNRESOLVED_IMPORT` browser errors
- React renders and state updates work correctly


JustInCache:
@hi-ogawa Can you please review ? Suggest feedback if any !! Thanks

packages/plugin-react/src/index.ts
```diff
@@ -172,7 +172,10 @@ export default function viteReact(opts: Options = {}): Plugin[] {
             {
               tag: 'script',
               attrs: { type: 'module' },
-              children: getPreambleCode(base),
+              // In bundled dev mode, Rolldown resolves module specifiers at build
+              // time without URL-level base stripping, so we must use '/' instead
+              // of config.base to match the resolveId hook for '/@react-refresh'.
```
sapphi-red:
```suggestion
              // In bundled dev mode, the src does not go through the middlewares
              // so we don't need to append the base
```

sapphi-red (approved):

PR #1200, opened 2026-04-24
chore: (experiment) invalidate virtual css on css change

hi-ogawa:
### Description

- Related to https://github.com/vitejs/vite-plugin-react/pull/1188#discussion_r3136149284

Experimenting on `collectCss + addWatchFile` technique. I hope Vite can make `addWatchFile` work, but if stretching this direction is too clumsy, we should just go back manual traverse and css invalidation approach like we had before https://github.com/vitejs/vite-plugin-react/pull/847

### DIFF
--- a/packages/plugin-rsc/src/plugin.ts
+++ b/packages/plugin-rsc/src/plugin.ts
@@ -2229,6 +2229,9 @@ function vitePluginRscCss(
     const visitedFiles = new Set<string>()
 
     function recurse(id: string) {
+      if (id.startsWith('\0virtual:vite-rsc/css?')) {
+        return
+      }
       if (visited.has(id)) {
         return
       }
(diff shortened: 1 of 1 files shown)


packages/plugin-rsc/src/plugin.ts
```diff
@@ -2229,6 +2229,9 @@ function vitePluginRscCss(
     const visitedFiles = new Set<string>()
 
     function recurse(id: string) {
+      if (id.startsWith('\0virtual:vite-rsc/css?')) {
+        return
+      }
```
hi-ogawa, line 2234: Not sure but I feel like we had this exact check at some point. Should go check blame.

hi-ogawa: Oh well, it looks like that's what I did for `?assets` query, but haven't gotten ported back https://github.com/hi-ogawa/vite-plugin-fullstack/blob/28e9540a68529c58842e9a3bf17d2193a065d524/src/plugin.ts#L563-L570

PR #1203, opened 2026-04-28
fix(rsc): convert vendored react-server-dom-webpack to ESM

### LINKED ISSUE #1204: Vendored react-server-dom-webpack CJS breaks on ESM-only runtimes (Cloudflare Workers)
`@vitejs/plugin-rsc` vendors `react-server-dom-webpack` as raw CJS into `dist/vendor/react-server-dom/`. These files use `require()` and `module.exports`.

On pure ESM runtimes (Cloudflare Workers, Deno Deploy), production builds can fail with `"require is not defined"` or `"exports is not defined"` when the CJS vendor code leaks into ESM chunks.

Reproduction: https://github.com/remorses/vite-rsc-cloudflare-cjs-repro

```bash
pnpm install
pnpm reproduce
```

The repro is based on the official `examples/starter-extra` Cloudflare setup and locks `@vitejs/plugin-rsc` to `0.5.21`. `pnpm reproduce` builds with the Cloudflare Vite plugin and fails because the RSC output still contains the CJS `react-server-dom-webpack-server.edge.production.js` bundle.

Fix in #1203.

remorses:
Fixes #1204

Reproduction: https://github.com/remorses/vite-rsc-cloudflare-cjs-repro

`@vitejs/plugin-rsc` vendors `react-server-dom-webpack` into `dist/vendor/` as raw CJS. These use `require()` and `module.exports`, which break on pure ESM runtimes (Cloudflare Workers, Deno Deploy).

`optimizeDeps` handles CJS-to-ESM conversion in dev mode, but it doesn't run during `vite build`. In production, Rolldown handles CJS files directly by wrapping them in `__commonJS` + a `__require` shim. This shim falls back to the real `require` global when its internal routing can't resolve the target. On Cloudflare Workers, `require` doesn't exist.

Shipping ESM vendor files eliminates the problem entirely since there's no CJS interop layer involved.

Adds an esbuild post-processing step in `tsdown.config.ts` that converts each CJS entry file to self-contained ESM with proper `import`/`export` statements. Removes the `cjs/` subdirectory after conversion. All 424 existing tests pass.

### DIFF
--- a/packages/plugin-rsc/package.json
+++ b/packages/plugin-rsc/package.json
@@ -59,6 +59,7 @@
     "@vitejs/plugin-react": "workspace:*",
     "@vitejs/test-dep-cjs-and-esm": "./test-dep/cjs-and-esm",
     "@vitejs/test-dep-cjs-falsy-primitive": "./test-dep/cjs-falsy-primitive",
+    "esbuild": "^0.28.0",
     "picocolors": "^1.1.1",
     "react": "^19.2.5",
     "react-dom": "^19.2.5",
--- a/packages/plugin-rsc/tsdown.config.ts
+++ b/packages/plugin-rsc/tsdown.config.ts
@@ -90,7 +90,6 @@ async function convertVendorToEsm(vendorDir: string) {
           'stream',
           'async_hooks',
         ],
-        define: { 'process.env.NODE_ENV': '"production"' },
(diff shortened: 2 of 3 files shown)


hi-ogawa:
Closing as the reproduction doesn't look right.

PR #1223, opened 2026-05-11
feat(plugin-rsc): expose RSC compatibility manifest

agcty:
## Summary

Adds a compiler-owned RSC compatibility manifest for frameworks built on `@vitejs/plugin-rsc`.

The manifest is intended for deployment skew protection. A framework can include `compatibilityManifest.compatibilityVersion` in RSC responses, headers, or framework manifests, then trigger its own recovery behavior when a request comes from an incompatible client.

## Usage

Framework runtime code can import the finalized manifest from RSC or SSR environments:

```ts
import compatibilityManifest from 'virtual:vite-rsc/compatibility-manifest'

export function getRscResponseMetadata() {
  return {
    compatibilityVersion: compatibilityManifest.compatibilityVersion,
  }
}
```

Frameworks with custom build orchestration can finalize the manifest through the plugin manager after the real RSC and client builds have completed:

```ts
const manager = getPluginApi(config)!.manager
const manifest = manager.finalizeCompatibilityManifest()
```

Build-time `manager.getCompatibilityManifest()` and `manager.getCompatibilityVersion()` throw until the manifest has been finalized, so scan builds or incomplete custom pipelines cannot accidentally emit a partial version.

## What The Manifest Covers

The compatibility version is derived from the build and RSC state owned by `@vitejs/plugin-rsc`:

- Vite base path
- RSC runtime package versions
- final assets manifest hash
- final client/RSC/SSR output bundle hashes
- client reference keys and rendered exports
- server reference keys and exported functions
- server-action encryption key identity when action closure encryption is emitted

Frameworks can compose this value with framework-specific deployment state such as route/data protocol versions, deployment ids, cache policy, or runtime-only configuration that plugin-rsc cannot observe directly.

### DIFF
--- a/packages/plugin-rsc/README.md
+++ b/packages/plugin-rsc/README.md
@@ -442,6 +442,46 @@ export default defineConfig({
 })
 ```
 
+### Framework compatibility manifest
+
+Frameworks can import `virtual:vite-rsc/compatibility-manifest` from the RSC or
+SSR environment to access compiler-owned deployment compatibility metadata. This
+is intended for deployment skew protection: a framework can include
+`compatibilityManifest.compatibilityVersion` in RSC responses and trigger a
+document reload when a later RSC request comes from an incompatible client.
+
+```js
+import compatibilityManifest from 'virtual:vite-rsc/compatibility-manifest'
+
+export function getRscResponseMetadata() {
+  return {
+    compatibilityVersion: compatibilityManifest.compatibilityVersion,
+  }
+}
+```
+
+The manifest includes the Vite base path, RSC runtime package versions, a hash
+of the final assets manifest, final output bundle hashes, client reference keys
+with rendered exports, server reference keys with exported functions, and a hash
+of the server-action encryption key when action closure encryption is actually
+emitted.
+
+Frameworks with a custom build pipeline can use `getPluginApi(config).manager`
+after the real RSC and client builds have completed:
+
+```js
+import { getPluginApi } from '@vitejs/plugin-rsc'
+
+const manager = getPluginApi(config).manager
+const manifest = manager.finalizeCompatibilityManifest()
+const version = manifest.compatibilityVersion
+```
+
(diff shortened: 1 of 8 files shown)


agcty:
@jacob-ebey @hi-ogawa I wanted to add some context for the motivation here.

This started from the React Router RSC skew issue and Jacob’s comment about possibly moving toward a more precise versioning system for RSC deployments:
https://github.com/remix-run/react-router/issues/14849#issuecomment-4103735228

The initial pass first tried solving skew protection in React Router/framework space by deriving a compatibility version from the framework’s RSC graph. After working through the edge cases, it seemed like the correctness boundary actually belongs lower, in `@vitejs/plugin-rsc`.

The plugin already owns the authoritative build-time state frameworks would otherwise have to approximate: client reference keys, rendered client exports, server references, the final assets manifest, final client/RSC/SSR outputs, and server-action encryption identity. If each framework reconstructs that separately, skew protection becomes fragile and likely divergent.

So this PR tries to expose a first-class, opt-in primitive:

- plugin-rsc provides the compiler/build compatibility manifest
- frameworks compose it with framework-specific state, like deployment id, route/data protocol version, cache policy, or runtime-only config
- frameworks still decide the recovery behavior: reload, refetch, tolerate the deployment, etc.

The intended runtime API is `virtual:vite-rsc/compatibility-manifest`, with `compatibilityManifest.compatibilityVersion` included in framework RSC responses/headers/manifests.

Would you consider this the right layer / boundary and shape for skew protection? Would love to hear your thoughts.

hi-ogawa:
> Would you consider this the right layer / boundary and shape for skew protection? Would love to hear your thoughts.

No, I don't think this particular patch is exposing the right data at right layer. For example, any vite plugin can access `base`, so framework / end-user integration can do that already. Other fields can be mostly the same since RSC plugin expose the entire `manager`. Please create an issue first to discuss the scope. Closing for now.

agcty:
@hi-ogawa fair point, created an issue!

PR #1226, opened 2026-05-14
fix(react-swc): don't apply React Refresh wrapper for non JSX files

### LINKED ISSUE #1225: `@vitejs/plugin-react-swc` Refresh wrapper erroneously added to non-component, error in Web Worker
### Related plugins

- [x] [plugin-react-swc](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react-swc)

- [ ] [plugin-rsc](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-rsc)

### Describe the bug

I have a react project that imports some type information from another package for use by a typescript service which is running in a web worker.

The file in question just exports one [Record<string, string>](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) where the key is the path to a `.d.ts` file and the value is the content of said file.

I expect this file to be treated as a regular file without the need for HMR or anything fancy - it's just data. This was all fine until I upgraded `@vitejs/plugin-react-swc` and started to see `ReferenceError: window is not defined` in the console. This led me through the inner workings of vite until I finally discovered this in the plugin:
```
const reactCompRE = /extends\s+(?:React\.)?(?:Pure)?Component/;
```
This regular expression was matching _within a string_ and causing the refresh wrapper to be added to the file and ultimately the script failing to load.

If the module was being used directly in the react app, this additional wrapping would cause no harm. However, because the module was being loaded within a web worker, the `window` object is not available and so you get the console error.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-1pqwywfy?file=src%2FApp.tsx

### Steps to reproduce

There is technically no error without using a web worker. I hope what I have is adequate to prove the flaw in the logic, though. If not, let me know and I'll see if I can add a basic web worker to the example.

The instructions are on the page, but I will repeat them here:

Observe the network traffic for the loading of two files. Despite both containing just strings, `example1.ts` is given a refresh wrapper but `example2.ts` is not.

This is because the former includes what appears to be a declaration of a react component. The plugin uses a regular expression that matches this, failing to detect that it is within a literal string.

This does not cause any problem here because they are being loaded within a window. However, if example1.ts was used by a web worker, loading it would fail with an error about window not being defined.

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (22) x64 Intel(R) Core(TM) Ultra 7 165H
    Memory: 10.97 GB / 31.45 GB
  Binaries:
    Node: 22.20.0 - C:\nvm4w\nodejs\node.EXE
    Yarn: 4.14.1 - C:\nvm4w\nodejs\yarn.CMD
    npm: 10.9.3 - C:\nvm4w\nodejs\npm.CMD
  Browsers:
    Chrome: 148.0.7778.97
    Edge: Chromium (146.0.3856.72)
    Firefox: 150.0.3 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.8115
```

### Used Package Manager

yarn

### Logs

Vite does not generate an error in the logs. The only evidence of the issue is in the browser console when it fails to load the affected module using a web worker.

### Validations

ArnaudBarre:
Fixes https://github.com/vitejs/vite-plugin-react/issues/1225

Given that SWC already enable fast refresh only for `jsx: true` and that since this plugin was not running on `.js`, this feel a good and safe change. This could be a breaking change in this unlikely set of circumstances:

- Using class component only
- Using a file format other than JSX/TSX, with pre-processing that generates `jsx_dev()`
- Configuring `parserConfig` to `(id) => id.endsWith(".ext") ? ({ syntax: 'ecmascript', jsx: false }) : null`

In that case hmr will go back to page reload. Changing to `jsx: true` will give back the previous behaviour

### DIFF
--- a/packages/plugin-react-swc/CHANGELOG.md
+++ b/packages/plugin-react-swc/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Don't apply React Refresh wrapper for non JSX files [#1226](https://github.com/vitejs/vite-plugin-react/pull/1226)
+
+Fixes [#1225](https://github.com/vitejs/vite-plugin-react/issues/1225).
+
 ## 4.3.1 (2026-05-14)
 
 ### Avoid `esbuild` warnings with Vite 8 [#1195](https://github.com/vitejs/vite-plugin-react/pull/1195)
--- a/packages/plugin-react-swc/src/index.ts
+++ b/packages/plugin-react-swc/src/index.ts
@@ -4,7 +4,6 @@ import { join } from 'node:path'
 import { exactRegex } from '@rolldown/pluginutils'
 import {
   type JscTarget,
-  type Output,
   type ParserConfig,
   type ReactConfig,
   type Options as SWCOptions,
@@ -197,15 +196,15 @@ const react = (_options?: Options): Plugin[] => {
           },
         )
         if (!result) return
-        if (!refresh) return result
+        if (!refresh || !result.jsx) return result.output
 
         const newCode = addRefreshWrapper(
-          result.code,
+          result.output.code,
           '@vitejs/plugin-react-swc',
           id,
           options.reactRefreshHost,
         )
-        return { code: newCode ?? result.code, map: result.map }
+        return { code: newCode ?? result.output.code, map: result.output.map }
       },
     },
     options.plugins || options.useAtYourOwnRisk_mutateSwcOptions
@@ -219,8 +218,8 @@ const react = (_options?: Options): Plugin[] => {
           configResolved(config) {
             viteCacheRoot = config.cacheDir
           },
-          transform: (code, _id) =>
-            transformWithOptions(
+          async transform(code, _id) {
+            const result = await transformWithOptions(
               _id.split('?')[0],
               code,
               'esnext',
@@ -230,7 +229,9 @@ const react = (_options?: Options): Plugin[] => {
                 runtime: 'automatic',
                 importSource: options.jsxImportSource,
               },
-            ),
+            )
+            return result?.output
+          },
         }
       : {
           name: 'vite:react-swc',
@@ -293,7 +294,6 @@ const transformWithOptions = async (
             : undefined
   if (!parser) return
 
-  let result: Output
   try {
     const swcOptions: SWCOptions = {
       filename: id,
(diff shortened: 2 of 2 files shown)


ArnaudBarre:
@sapphi-red I'm wondering if for the [rolldown builtin refresh wrapper](https://github.com/rolldown/rolldown/blob/ffb8abd7bdf2be9326cb33001af6d425aa05fef5/crates/rolldown_plugin_vite_react_refresh_wrapper/src/lib.rs#L52) you access the ast and do something like:

```ts
let only_react_comp =
  !has_refresh &&
  // pseudo JS code
  ast.body.statements.some(
    (s) =>
      s.kind === 'ClassDeclaration' &&
      s.extends.name.match(/^(?:React\.)?(?:Pure)?Component$/),
  )
```

sapphi-red (approved):

sapphi-red:
I guess we can also fix the error by changing this `window` with `globalThis`:
https://github.com/vitejs/vite-plugin-react/blob/5764f967e0166cb542a0486f2e013eba95d4613d/packages/common/refresh-runtime.js#L599

sapphi-red:
> @sapphi-red I'm wondering if for the [rolldown builtin refresh wrapper](https://github.com/rolldown/rolldown/blob/ffb8abd7bdf2be9326cb33001af6d425aa05fef5/crates/rolldown_plugin_vite_react_refresh_wrapper/src/lib.rs#L52) you access the ast and do something like:

It doesn't have access to the AST here since the `transform` hook is string-in-string-out.

ArnaudBarre:
> I guess we can also fix the error by changing this window with globalThis

For the past years I've found benefits to keep using window there because it allowed to find cases with bad configuration or people over bundling files in their workers. The idea was also that if one day we figure out HMR for workers, it will not be a breaking change currently the client can't run there.

But maybe the chance is so low that we get there that we should make it no-op to load the HMR client in a worker

ArnaudBarre:
> It doesn't have access to the AST here since the transform hook is string-in-string-out
And I suppose it would cost to much memory to pass it around via some kind of context?

sapphi-red:
> > It doesn't have access to the AST here since the transform hook is string-in-string-out
>
> And I suppose it would cost to much memory to pass it around via some kind of context?

yeah, passing it around would probably cost than parsing it.

ArnaudBarre:
@sapphi-red should we merge this?

PR #1234, opened 2026-05-21
fix(rsc): handle `export *` re-exports in `use client` and `use server` modules

### LINKED ISSUE #1233: [plugin-rsc] `use-client` transform throws on `export * from` (ExportAllDeclaration)
## Describe the bug

`@vitejs/plugin-rsc`'s `rsc:use-client` transform throws

```
Error: unsupported ExportAllDeclaration
    at transformProxyExport (.../plugin-rsc/dist/transforms/index.js:338:90)
```

for any `"use client"` module that re-exports another module with `export * from '...'` or `export * as Ns from '...'`. This is a common pattern in real-world React code (barrel files for groups of client components), and it ships in the Next.js `rsc-basic` e2e suite ([components/export-all/](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/rsc-basic/components/export-all/index.js)), so any Vite/RSC framework that aims to be a drop-in replacement for the Next.js App Router build pipeline hits this immediately.

## Reproduction

A `"use client"` file containing:

```js
"use client";
export * from "./one";
```

triggers the error during the RSC environment build.

## Source

The throw is unconditional once `ignoreExportAllDeclaration` is false (which is the default for the use-client path):

```js
// dist/transforms/index.js around line 338
if (!options.ignoreExportAllDeclaration && node.type === "ExportAllDeclaration")
  throw new Error("unsupported ExportAllDeclaration");
```

There is no public option on the use-client plugin to flip this, and the `*` case isn't routed through the proxy generator.

## Expected behavior

`export * from '...'` and `export * as Ns from '...'` should be handled inside a `"use client"` boundary. Two reasonable options:

1. **Resolve at transform time** — enumerate the target module's named exports and emit one client-reference proxy per name (skipping `default`, matching ESM `export *` semantics).
2. **Forward at module-graph time** — leave the declaration in place and let the bundler resolve it natively after the proxies for the target module are emitted. This is conceptually what Next.js's flight loader does: it falls back to `assumedSourceType === "module"` when `clientRefs` contains `*` so webpack handles the re-export ([next-flight-loader/index.ts](https://github.com/vercel/next.js/blob/canary/packages/next/src/build/webpack/loaders/next-flight-loader/index.ts#L25-L54)).

## Current workaround

In cloudflare/vinext we added a pre-pass plugin that runs before `rsc:use-client` in the RSC environment, parses each `"use client"` module, and rewrites:

- `export * from '...'` → `export { a, b, c, ... } from '...'` (named exports enumerated by resolving the target via `this.resolve` + `this.load`, recursing through chained `export *`)
- `export * as Ns from '...'` → `import * as Ns from '...'; export { Ns };`

Tracking PR: <https://github.com/cloudflare/vinext/pull/1373>
Plugin source: <https://github.com/cloudflare/vinext/blob/fix/issue-1352-use-client-export-all/packages/vinext/src/plugins/use-client-export-all.ts>

This works for relative imports inside the user's graph but cannot handle bare-specifier targets that the RSC plugin's package-source path will eventually proxy — those still hit the error.

## System info

`@vitejs/plugin-rsc@0.5.26` on Vite 8 / Rolldown.

james-elicx:
## Summary

The `use-client` proxy transform in `@vitejs/plugin-rsc` throws `unsupported ExportAllDeclaration` whenever a `"use client"` file contains `export * from '...'`. Reported downstream in [cloudflare/vinext#1352](https://github.com/cloudflare/vinext/issues/1352), where Next.js's app-dir test suite hits this with a real-world `components/export-all/index.js` fixture.

fixes #1233

## Reproduction

A new fixture under `packages/plugin-rsc/examples/basic/src/routes/export-all/` reproduces the original error on `main`:

```
[plugin rsc:use-client] .../routes/export-all/index.tsx
Error: unsupported ExportAllDeclaration
    at transformProxyExport (.../transforms/index.js:338:90)
```

It builds cleanly with the fix.

## Fix

1. **Pure transforms** (`proxy-export.ts`, `wrap-export.ts`) now handle the `export * as ns from '...'` form directly, since the namespace name is statically known.
2. **Plugin layer** (`plugin.ts`) introduces an `expandExportAllDeclarations` helper used by both `vitePluginUseClient` and `vitePluginUseServer`. For each bare `export * from '...'` declaration, it resolves the source, recursively collects the named exports via AST walk, and rewrites the source to `export { a, b, c } from '...'` before the proxy transform runs. This mirrors what Next.js's webpack plugin does to pre-resolve `export *` into named exports.

   - In **build** mode it uses `this.load({ id }).code` (which Rollup populates).
   - In **dev** mode `ModuleInfo.code` is unsupported and `transformRequest` returns module-runner output (`__vite_ssr_exportName__(...)`) that lacks readable ESM exports, so it falls back to `fs.readFile` + `transformWithOxc` to get standard JS the AST walk can read.

## Test plan

- [x] New unit coverage in `proxy-export.test.ts` for `export * as ns from`, the post-resolution form, the `ignoreExportAllDeclaration` escape hatch, and the still-thrown unresolved-bare case; updated `wrap-export.test.ts` snapshot. All 428 unit tests pass.
- [x] New e2e test `export *` in `basic.test.ts` against the new fixture. All 151 basic e2e tests pass in both dev and build modes.
- [x] Verified original build error reproduces on `main` (stashed the fix, rebuilt, build fails with the exact error from the issue) and disappears with the fix.

### DIFF
--- a/packages/plugin-rsc/examples/basic/src/routes/export-all/index.tsx
+++ b/packages/plugin-rsc/examples/basic/src/routes/export-all/index.tsx
@@ -1,0 +1,7 @@
+'use client'
+
+export * from './named'
+
+export function ExportAllNamed() {
+  return <span data-testid="export-all-named">export-all-named</span>
+}
--- a/packages/plugin-rsc/src/plugin.ts
+++ b/packages/plugin-rsc/src/plugin.ts
@@ -1413,20 +1413,16 @@ async function collectExportNames(
   if (seen.has(resolvedId)) return []
   seen.add(resolvedId)
 
-  // `this.load`'s ModuleInfo.code is build-only. In dev mode, the dev
-  // environment's `transformRequest` returns module-runner-specific output
-  // (e.g. `__vite_ssr_exportName__("X", ...)` instead of `export ... from`),
-  // which the AST walk below can't read. Use a plain TS/JSX-aware transform
-  // on the source instead so we get standard ESM exports to walk.
+  // Read the source from disk and strip TS/JSX so the AST walk below sees
+  // standard ESM exports. We don't go through `this.load` /
+  // `transformRequest` here — in dev they return module-runner output
+  // (`__vite_ssr_exportName__(...)`) the walker can't read, and on build
+  // there's no practical benefit over reading the source directly for the
+  // simple TS/JSX modules we care about.
   let moduleCode: string | undefined
   try {
-    if (ctx.environment.mode === 'dev') {
-      const raw = await fs.promises.readFile(resolvedId, 'utf-8')
-      moduleCode = await transformSourceForExportScan(raw, resolvedId)
-    } else {
-      const moduleInfo = await ctx.load({ id: resolvedId })
-      moduleCode = moduleInfo.code ?? undefined
-    }
+    const raw = await fs.promises.readFile(resolvedId, 'utf-8')
+    moduleCode = await transformSourceForExportScan(raw, resolvedId)
   } catch {
     return []
   }
--- a/packages/plugin-rsc/src/plugin.ts
+++ b/packages/plugin-rsc/src/plugin.ts
@@ -23,7 +23,6 @@ import {
   isCSSRequest,
   normalizePath,
   parseAstAsync,
-  transformWithOxc,
 } from 'vite'
 import { crawlFrameworkPkgs } from 'vitefu'
 import vitePluginRscCore from './core/plugin'
@@ -1377,6 +1376,30 @@ function globalAsyncLocalStoragePlugin(): Plugin[] {
   ]
 }
 
+// Strip TS/JSX so `parseAstAsync` can read the result. Prefer oxc when
+// available (Vite 8+); fall back to esbuild for older Vite versions.
+async function transformSourceForExportScan(
+  code: string,
+  filename: string,
+): Promise<string | undefined> {
+  const v = vite as Partial<{
+    transformWithOxc: (
+      code: string,
+      filename: string,
+      options?: { sourcemap?: boolean },
+    ) => Promise<{ code: string }>
+    transformWithEsbuild: (
+      code: string,
+      filename: string,
+      options?: { sourcemap?: boolean },
+    ) => Promise<{ code: string }>
+  }>
+  const transform = v.transformWithOxc ?? v.transformWithEsbuild
+  if (!transform) return undefined
+  const result = await transform(code, filename, { sourcemap: false })
+  return result.code
+}
+
 // Recursively collect the named exports of a module (following `export * from`
 // chains), so that the RSC `use client`/`use server` proxy transforms can
 // expand bare `export *` re-exports into explicit named re-exports before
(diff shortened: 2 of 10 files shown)


packages/plugin-rsc/src/plugin.ts
```diff
@@ -1375,6 +1376,147 @@ function globalAsyncLocalStoragePlugin(): Plugin[] {
...
+  // which the AST walk below can't read. Use a plain TS/JSX-aware transform
+  // on the source instead so we get standard ESM exports to walk.
+  let moduleCode: string | undefined
+  try {
+    if (ctx.environment.mode === 'dev') {
+      const raw = await fs.promises.readFile(resolvedId, 'utf-8')
+      moduleCode = await transformSourceForExportScan(raw, resolvedId)
+    } else {
+      const moduleInfo = await ctx.load({ id: resolvedId })
+      moduleCode = moduleInfo.code ?? undefined
```
hi-ogawa: I don't think switching to `this.load` on build is practically meaningful. Can we start with minimal js/tsx support?

hi-ogawa (commented):
Thanks for working on this. The idea makes sense to me.

packages/plugin-rsc/src/plugin.ts
```diff
@@ -1375,6 +1376,143 @@ function globalAsyncLocalStoragePlugin(): Plugin[] {
...
+        if (subResolved) {
+          names.push(...(await collectExportNames(ctx, subResolved.id, seen)))
+        }
+      }
+    }
+  }
+  return names
+}
+
+async function expandExportAllDeclarations(
```
hi-ogawa, line 1479: probably it's better to extract this as `src/transforms/` util with unit test.

packages/plugin-rsc/examples/basic/src/routes/export-all/index.tsx
```diff
@@ -0,0 +1,7 @@
+'use client'
+
+export * from './named'
```
hi-ogawa, line 3: coverage for `"use server"` too

packages/plugin-rsc/src/plugin.ts
```diff
@@ -1375,6 +1376,143 @@ function globalAsyncLocalStoragePlugin(): Plugin[] {
...
+    return []
+  }
+  if (!moduleCode) return []
+
+  let ast: Awaited<ReturnType<typeof parseAstAsync>>
+  try {
+    ast = await parseAstAsync(moduleCode)
+  } catch {
+    return []
+  }
```
hi-ogawa, line 1436: we can start from surfacing these errors to start with so it's easier to iterate on unsupported case.

hi-ogawa (approved):
Thanks for the follow-up. I reviewed this again, and it looks good to merge.

I left a few comments for things I might clean up separately on my side. Mainly I want to extract `expandExportAllDeclarations` into a pure-ish utility so we can unit test it directly. There are also a few edge cases I want to tweak for export name conflicts and load/parse failure mode. In case interested, here is my follow-up note: https://gist.github.com/hi-ogawa/cf434aff19cc869e6cfae366851db0c7

packages/plugin-rsc/src/transforms/wrap-export.ts
```diff
@@ -168,18 +168,25 @@ export function transformWrapExport(
...
     // for now we just give an option to not throw for this case.
     // https://github.com/vitejs/vite-plugin-vue/blob/30a97c1ddbdfb0e23b7dc14a1d2fb609668b9987/packages/plugin-vue/src/main.ts#L372
-    if (
-      !options.ignoreExportAllDeclaration &&
-      node.type === 'ExportAllDeclaration'
-    ) {
-      throw Object.assign(new Error('unsupported ExportAllDeclaration'), {
-        pos: node.start,
-      })
+    if (node.type === 'ExportAllDeclaration') {
```
hi-ogawa, line 177: also, this changes looks unneeded as it cannot be useful for server function case at least.

PR #1239, opened 2026-05-29
fix(rsc): handle conflict names from `export *` in `use client` and `use server` modules

hi-ogawa:
### Description

- Follow up to https://github.com/vitejs/vite-plugin-react/pull/1234
- review plan https://gist.github.com/hi-ogawa/cf434aff19cc869e6cfae366851db0c7

This PR extracts `export *` expansion into a dedicated transform with file fixtures based unti tests and tightens its handling of ESM star-export ambiguity before downstream proxy/wrap transforms run.

It turned out `export *` semantics have many edge cases and some cases are explicitly not handled yet but kept as test cases with TODOs, for example:

- Support duplicate `export *` entries that resolve to the same underlying binding instead of treating them as ambiguous.
- Support duplicate namespace re-exports that resolve to the same namespace binding.
- Replace the current cycle bailout with per-name/per-binding resolution so cyclic star exports that resolve to the same binding can survive.
- Support string-literal export names instead of rejecting them.
- Preserve/compose source maps for the pre-rewrite transform.


hi-ogawa:
@codex review

packages/plugin-rsc/src/transforms/wrap-export.ts
```diff
@@ -175,14 +185,7 @@ export function transformWrapExport(
...
     if (node.type === 'ExportAllDeclaration') {
-      if (node.exported?.type === 'Identifier') {
-        tinyassert(node.source.type === 'Literal')
-        const exportedName = node.exported.name
-        const localName = `$$import_${exportedName}`
-        output.remove(node.start, node.end)
-        toAppend.push(`import * as ${localName} from ${node.source.raw}`)
-        wrapExport(localName, exportedName)
-      } else if (!options.ignoreExportAllDeclaration) {
+      if (!options.ignoreExportAllDeclaration) {
```
hi-ogawa, line 188: This was reverted since this change was irrelevant for original PR
- https://github.com/vitejs/vite-plugin-react/pull/1234#discussion_r3322795175

PR #1251, opened 2026-06-15
fix(rsc): preserve server reference metadata across environments

### LINKED ISSUE #1250: Inline "use server" actions can be missing from build server-references manifest
### Related plugins

### Describe the bug

An inline `"use server"` action defined in a module with **no top-of-file `"use server"` directive** can be dropped from the production server-references manifest in build flows where the manifest is emitted after non-rsc passes sweep the shared metadata map.

Dev is unaffected, and the default build ordering usually masks this because the real rsc build can re-register the entry before `virtual:vite-rsc/server-references` is emitted. The bug becomes observable in separate-pass prerender/cache pipelines: the app builds and prerenders successfully, but the cached Flight payload later revives the action from its id alone. When that action is invoked, runtime resolution goes through the shipped manifest; if the entry is missing, it throws **`server reference not found`**.

Root cause: `rsc:use-server` runs once per environment, and all environments share one `serverReferenceMetaMap` (the map the manifest is generated from). The rsc transform registers the inline action; the ssr/client transform then deletes it because the module has no file-level directive. That `delete manager.serverReferenceMetaMap[id]` is valid dev HMR cleanup for removed file-level directives, but it should not run during build, where the shared map feeds the production manifest.

I have a fix ready: scope the non-rsc delete to dev so build manifests keep inline action entries.

### Reproduction

https://github.com/ivogt/vite-plugin-react/tree/fix/inline-server-reference-build-manifest

### Steps to reproduce

  Note: a vite.new app usually does NOT surface this because the default build order masks it  (the rsc build can re-register before the manifest is emitted). The minimal deterministic  reproduction is the unit test in the PR: it exercises the real non-rsc cleanup branch and  manifest generator, and shows `export default {};` when the inline entry is dropped. A runtime  reproduction needs a separate-pass prerender/cache pipeline where the shipped manifest is emitted  after the non-rsc sweep. Link the PR (or branch) here once opened.

1. Define an inline `"use server"` action inside a component, in a module with no top-of-file `"use server"`.
2. Render that component and capture the RSC Flight output (prerender or cache it).
3. Emit the production server-references manifest after the ssr/client transform has swept the shared metadata map, with no later rsc transform re-adding the entry.
4. Serve the stored payload on a later request **without re-running the component**.
5. Invoke the action → `server reference not found`.

https://github.com/ivogt/vite-plugin-react/tree/fix/inline-server-reference-build-manifest 
Or, at the unit level: run `packages/plugin-rsc/src/use-server-inline-action.test.ts` with the non-rsc delete left unconditional — the generated manifest comes out `export default {};` and the build assertion fails.

### System Info

```shell
System:
  OS: macOS 26.5
  CPU: (10) arm64 Apple M4
  Memory: 239.78 MB / 24.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.12.0 - /Users/ivotodorov/.nvm/versions/node/v24.12.0/bin/node
  npm: 11.6.2 - /Users/ivotodorov/.nvm/versions/node/v24.12.0/bin/npm
  pnpm: 10.34.1 - /Users/ivotodorov/.nvm/versions/node/v24.12.0/bin/pnpm
  bun: 1.3.9 - /Users/ivotodorov/.bun/bin/bun
Browsers:
  Chrome: 149.0.7827.103
  Safari: 26.5
npmPackages:
  @vitejs/release-scripts: ^1.7.1 => 1.7.1
  vite: ^8.0.16 => 8.0.16
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

ivogt:
Fixes #1250.

An inline `"use server"` action is registered in the shared `serverReferenceMetaMap` by the RSC environment. When the same module also flows through an SSR or client graph, that non-RSC transform could delete the RSC-owned entry.

This causes two user-visible failures:

- **Dev:** submitting the inline-action form fails with `[vite-rsc] invalid server reference` because reference validation reads the same metadata map.
- **Build:** the action is omitted from the production server-references manifest. Replaying persisted Flight without rerunning the component then fails on invocation with `server reference not found`.

## Concrete Scenario

`examples/prerender-inline-action` demonstrates a framework-style route module that exports both static metadata and a component containing an inline server action. The SSR entry consumes only the metadata, while the component renders exclusively in the RSC environment.

Its production pipeline runs separate passes:

1. Build and execute an RSC prerender entry, writing `dist/flight/page.flight`.
2. Analyze the SSR graph, which imports the route metadata.
3. Build a runtime-only RSC entry that replays the stored Flight payload without importing or rerunning the page component.
4. Build the client and final SSR bundles.

Because the revived action contains only its reference ID, invocation depends on the server-references manifest retaining the RSC-owned metadata.

## Fix

Server-reference metadata now records the environment that registered it. An entry may be deleted only by the RSC environment, which is authoritative for server references, or by the environment that registered the entry:

```ts
const deleteServerReferenceMeta = (environmentName: string, id: string) => {
  const meta = manager.serverReferenceMetaMap[id]
  if (!meta) return
  if (
    environmentName === serverEnvironmentName ||
    environmentName === meta.environmentName
  ) {
    delete manager.serverReferenceMetaMap[id]
  }
}
```

Cleanup behavior remains intact:

- The RSC environment still removes stale entries when an inline or file-level directive is removed.
- Non-RSC environments still remove their own file-level proxy registrations.
- Non-RSC transforms no longer delete entries owned by the RSC environment.

## Tests

- `e2e/prerender-inline-action.test.ts` covers both dev and production. It invokes the action from `/cache`, restarts the server, replays the persisted payload at `/replay`, and invokes the action again without rerunning its component.
- Without the fix, the dedicated test fails in dev with `invalid server reference` and in production with `server reference not found`.
- `src/use-server-meta-cleanup.test.ts` covers metadata ownership and stale-entry cleanup semantics.
- The original `e2e/cache-replay.test.ts` and `examples/cache-replay` remain unchanged.

Verified locally:

- Dedicated E2E: 2 passed, covering dev and build.
- Original cache-replay E2E: 2 passed, covering dev and build.
- Unit tests: 452 passed.
- Typecheck, lint, and format: passed.

## Related

- Issue: #1250
- #753 introduced cleanup for removed file-level directives; this PR preserves that cleanup through metadata ownership.


ivogt:
I updated the PR description with the concrete cache/prerender scenario and why runtime re-registration cannot recover this: on a hit, Flight only has id + bound args, so the production server-references manifest is the only path back to the module.



Here is the concrete cache/prerender shape that makes the bug observable:

```tsx
// article-list.tsx
// No top-of-file "use server"

export function ArticleList() {
  const articleId = "post-1";

  async function like() {
    "use server";
    await db.like(articleId);
  }

  return (
    <form action={like}>
      <button>Like</button>
    </form>
  );
}
```

```ts
// prerender/cache write
const stream = await renderToReadableStream(<ArticleList />);
const flightString = await new Response(stream).text();

rscStringStore.set("/articles", flightString);
```

```ts
// later request: cache/prerender hit
const flightString = rscStringStore.get("/articles");

// ArticleList is NOT re-run here.
// The inline `like` body does NOT execute.
// React revives only a server reference from the stored Flight payload:
//   id: "5f9c124a333c#$$hoist_0_like"
//   bound args: ["post-1"]
return new Response(flightString, {
  headers: { "content-type": "text/x-component" },
});
```

```ts
// later action POST
await loadServerAction("5f9c124a333c#$$hoist_0_like");

// plugin-rsc effectively does:
const [file, name] = id.split("#");
// file = "5f9c124a333c"

const mod = await requireModule(file);
// in production, requireModule resolves through:
// serverReferences["5f9c124a333c"]()
```

If `virtual:vite-rsc/server-references` was emitted after the non-RSC transform deleted the inline entry, the shipped manifest is effectively:

```ts
export default {};
```

So runtime resolution fails here:

```ts
serverReferences["5f9c124a333c"]; // undefined
```

with:

```txt
Error: server reference not found '5f9c124a333c'
```

The stored Flight string is valid: it contains the server reference id and bound args. What is missing is the production manifest entry that maps the id back to the module. Since the cache/prerender hit never re-runs `ArticleList`, there is no later opportunity for `registerServerReference(...)` to repair the missing manifest entry.

hi-ogawa (changes requested):
Thanks for providing the details. I've taken a look, and I don't think the proposed fix and accompanying tests establish that the current behavior is a bug in the RSC plugin.

The provided tests seem to simulate files with inline server actions passing through a non-RSC environment module graph, but this is not an expected use case in the RSC model.

Inline actions belong to RSC modules. Non-RSC environments may consume file-level `"use server"` modules because those become proxies, but directly including an inline-action component in SSR/client appears invalid. This should ideally be validated by the RSC plugin and surfaced as a build error, which we do not currently do.

Please provide a concrete use case, ideally with complete application or framework code, that demonstrates the issue. It should then be ported to an E2E test so I can diagnose it properly. Please also review the new [contributor testing guide](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-rsc/CONTRIBUTING.md) for guidance on the desired type of testing.

ivogt:
Hi @hi-ogawa , 
Thank you for looking into this pr and for supporting the plugin!
I've updated and refactored test and now cache-replay is extended to do pre-render in separate Vite step. 

I've also refactored the fix and now it is essentially only preventing other envs (like ssr) to clear the map. - https://github.com/vitejs/vite-plugin-react/pull/1251/changes#diff-a07541c27f70722e3fd53dc27c03e8b1b2cdbc2d789e1896d1ef8cbd21e9895dR2011 

The test app prerenders in a separate build pass of the RSC environment: buildApp builds a prerender entry that renders the pages declared in src/inline-pages.tsx and writes them into Flight files under dist/flight/, then builds the runtime-only RSC entry that replays them without importing the page modules. 
Ref: entry.prerender.rsc.tsx  - https://github.com/vitejs/vite-plugin-react/pull/1251/changes#diff-bf0572bbd0523fb75c7510ca2bffb74ed062792de20210ef834f9f50b586d129R11 
Ref: https://github.com/vitejs/vite-plugin-react/pull/1251/changes#diff-15f2199108f4e39796f72ee95f6b220a51190a692b975f4cd16c04357f23bee5R9


Happy to further improve this fix if needed.

Thank you.

hi-ogawa (changes requested):
Instead of updating existing example, can you add a dedicated example?

ivogt:
@hi-ogawa Moved this into a dedicated example. I considered whether the name might be too specific, but prerender-inline-action works well for E2E coverage because any failure will be immediately identifiable in CI.

hi-ogawa:
Thanks for the updates. But this still relies on the same pattern I flagged earlier https://github.com/vitejs/vite-plugin-react/pull/1251#pullrequestreview-4721586560.

In the example, `entry.ssr.tsx` imports `pageTitle` from `inline-page.tsx`, which is the module holding the inline `"use server"` action. That pulls an inline-action module into the SSR graph, which isn't expected in the RSC model. 

I'm going to close this and the issue for now, but happy to reopen if there's a real framework/app case where an inline-action module unavoidably ends up in a non-RSC graph.

hi-ogawa:
Just found that the missing validation isn't intended and I've re-opened https://github.com/vitejs/vite-plugin-react/issues/883#issuecomment-5029243311

PR #1252, opened 2026-06-17
feat(rsc): split server and client runtimes

TheAlexLichter:
This PR splits the combined RSC runtime into dedicated server and client entry
points.

Framework RSC entries that only render Flight payloads, decode actions, and register references can now use `@vitejs/plugin-rsc/rsc/server`. This avoids evaluating `client.edge` during startup and results in a smaller module graph.

The existing combined entry is still available for code that needs both sides.

Encrypted action modules can benefit too. The Flight client decoder is now loaded only when encrypted bound arguments are actually decrypted.

## Framework scope

- ✅ **React Router** can switch its RSC entry directly.
- ✅ **TanStack Start** can switch after removing its unused RSC-side `createFromReadableStream` re-export.
- ✅ **vinext** can switch its main RSC entry. When using `"use cache"`, it still loads the client protocol.
- ❌ **Waku** still needs the client runtime in its current combined RSC handler.

So the improvement only applies when the startup module graph does not load the Flight client through another path.

### Dev

Using vinext's [generated 33-route app](https://github.com/cloudflare/vinext/blob/d077d67b79a302c1cc75aa9eb333f3ea3a174936/benchmarks/generate-app.mjs), extended with cache and encrypted-action test routes, **cold dev startup** changed as follows:

| Scenario               |  Combined |     Split | Difference |
| ---------------------- | --------: | --------: | ---------: |
| Plain request          | 1451.1 ms | 1367.9 ms |   -83.2 ms |
| Encrypted bound action | 1439.3 ms | 1374.1 ms |   -65.2 ms |
| `"use cache"`          | 1476.5 ms | 1474.0 ms |   -2.5  ms |

This goes in line with [my experiments in vinext](https://github.com/cloudflare/vinext/pull/2080).

### Prod

Lazy loading moves a small amount of work to the first encrypted action.
In a **production** `vp preview` benchmark on my machine, the first action was 0.9 ms slower. 

I did not observe a throughput or steady-state latency regression.


hi-ogawa:
The PR description on Why may not be accurate from what I currently understand. The split might be still reasonable but we probably want to evaluate the actual effect of the change properly. Here is the review re-narrated by AI:

- The perf win seems narrower than the PR description suggests. Server action closure encryption still needs the Flight client because `utils/encryption-runtime` imports `createFromReadableStream` from `react/rsc/client`. So apps using encrypted bound args will load `client.edge` anyway, just not through the framework entry.
- Some listed beneficiaries may also need the client half independently. For example, Waku and vinext use RSC stream deserialization/caching paths at the framework layer, which likely means they still import/use `createFromReadableStream`.
- The split still seems reasonable as API shape. A framework that only does render/action decoding and does not need encrypted closure args or RSC stream deserialization can now avoid `client.edge`. TanStack Start may currently fit that category.
- Since vinext is cited as motivation, can we include/verify the benchmark numbers for that specific case? It would be useful to know whether the improvement still holds with realistic settings, especially with server action encryption and any framework-level RSC caching enabled.
- Even if the benchmark improves, this may be optimizing a narrower cold-start path rather than the general RSC runtime cost. The PR description should probably qualify that.

hi-ogawa (changes requested):
see above comment

TheAlexLichter:
@hi-ogawa thanks for the review! I've:

* Updated the PR description to address the feedback
* Lazy loaded the flight decoder to make sure most frameworks can benefit from the change

hi-ogawa (approved):
Thank you!

PR #1253, opened 2026-06-18
feat(rsc): ability to preserve decoded server references

james-elicx:
Extracts the core, React, and server-manifest reference-preservation changes from #1246.

---

RSC frameworks may need to cache, proxy, or compose Flight fragments before sending them back to the browser. A concrete example is a `use cache` wrapper that stores a rendered product card as a Flight stream and later replays it into a new RSC response.

That cached fragment can still contain a server action, like an “add to cart” action owned by the framework/runtime. During replay, the cache layer should not have to import the action implementation from the app bundler manifest; it only needs to keep the action reference opaque so the browser can call it through the normal action path later.

This adds an opt-in `serverReferences: "preserve"` decode mode for that case. It also preserves and deduplicates server-reference export names when transforms share a module, and adds an experimental `serverReferenceMarkers` option so preceding transforms can mark modules whose server-reference metadata must survive the `"use server"` HMR cleanup pass.

The e2e caches a Flight stream containing a framework-owned server action reference, replays that cached stream, and verifies the replayed client UI can still invoke the action.

### DIFF
--- a/.github/workflows/ci-rsc.yml
+++ b/.github/workflows/ci-rsc.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
         with:
           persist-credentials: false
-      - uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
+      - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
           node-version: 24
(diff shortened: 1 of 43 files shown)


hi-ogawa:
I've merged a PR https://github.com/vitejs/vite-plugin-react/pull/1289 to allow cache replay without importing server reference modules.

I will try to understand the idea of `serverReferenceMarkers` and `frameworkServerReferences`. I'd appreciate if you can craft e2e (dedicated example like I did in https://github.com/vitejs/vite-plugin-react/pull/1289 would be also welcome), so the motivation is clear and also can exercise more code pass. From my reading, current `frameworkServerReferences` in this PR likely doesn't work with progressive form action usage.

hi-ogawa (changes requested):
Labeling requested changes for the above ground.

james-elicx:
Updated the vinext pr based on #1289 + my other open prs - https://github.com/cloudflare/vinext/pull/2156/changes

PR #1254, opened 2026-06-18
fix(rsc): fix `transformWrapExport ` with `filter` and `rejectNonAsyncFunction`

james-elicx:
_Updated description by @hi-ogawa_

`transformWrapExport` currently validates `filter`-ed exports and includes them in `exportNames` even though their values are not wrapped. It also makes metadata exports fail async validation and reports references that were never registered. 

This behavior wasn't a problem for rsc plugin's internal use for rsc css transform https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-rsc/README.md#css-support, but is a problem for more general usages such as the integration indicated in  #1246.

This PR fixes `transformWrapExport` so that `filter` and `rejectNonAsyncFunction` options and `exportNames` return value behaves naturally and consistently. It also refines `ExportMeta.isFunction: boolean | undefined` detection so `filter` can make use of it more flexibly. The new test is added to cover hopefully all possible patterns of `ExportMeta`.

---

_Original description_

Extracted from #1246.

`transformWrapExport` currently validates filtered exports and includes them in `exportNames` even though their values are not wrapped. This makes metadata exports fail async validation and reports references that were never registered.

This applies filtering before validation/reporting and keeps unresolved identifier exports eligible for wrapping. Tests cover named exports, default exports, and unknown identifier metadata.


hi-ogawa (approved):
Hey, thanks for the fix! This looks like a nicely scoped change, so I've adjusted a bit and will be merging. Will try to capture larger picture as I go through other PRs.

PR #1255, opened 2026-06-18
feat(rsc): support inline server action methods

james-elicx:
Extracted from #1246.

Inline `"use server"` transforms currently produce invalid output for object methods and static class methods because method syntax cannot be replaced like a standalone function expression.

This rewrites supported methods to property/field assignments and gives their hoists method-derived names. Tests cover identifier and computed object/static methods, plus explicit errors for instance, private, getter, and setter methods.


hi-ogawa:
This isn't resolved yet. Now tracked in https://github.com/vitejs/vite-plugin-react/issues/1335.

PR #1256, opened 2026-06-18
fix(rsc): reset stateful directive regexes

james-elicx:
Extracted from #1246.

Global and sticky directive regexes retain `lastIndex`, so later functions or repeated transforms can miss matching directives.

This resets regex state around each match. Tests cover both `g` and `y` regexes, multiple functions, repeated transforms, and the final `lastIndex`.


hi-ogawa (changes requested):
Do you have a use case for `y`? The `g` case works without resetting.

The fix seems correct and don't mind including it but I'm not sure if when lib provides regex config/option like this, it's always lib side responsibility to normalize in general.

james-elicx:
It was a defensive change in the original PR. Likely not necessarily in reality. Sorry.

PR #1257, opened 2026-06-18
feat(rsc): support stable inline action names

james-elicx:
Extracted from #1246.

Inline action hoist names are currently based on traversal order, so inserting an unrelated action renumbers existing server-reference IDs.

This adds an opt-in `stableName` transform option using a content-derived hash with duplicate disambiguation. Tests verify stability across unrelated insertions, changes when the function changes, and deterministic duplicate handling.

### DIFF
--- a/packages/plugin-rsc/src/transforms/hoist.test.ts
+++ b/packages/plugin-rsc/src/transforms/hoist.test.ts
@@ -90,6 +90,16 @@ describe(transformHoistInlineDirective, () => {
     return result.names
   }
 
+  async function testStableTransformNames(input: string) {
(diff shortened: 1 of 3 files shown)


hi-ogawa:
I inferred the high level motivation of this PR as a following issue:
- https://github.com/vitejs/vite-plugin-react/issues/1395

I don't think hoisted function name is an actual requirement for this topic. Instead, the framework can generate a proper cache key and coordinate invalidation. Haven't fully fleshed out but I will try to illustrate the idea in https://github.com/vitejs/vite-plugin-react/pull/1411.

Please let me know if I misread the context.

PR #1258, opened 2026-06-18
feat(rsc): expose server function metadata

james-elicx:
Extracted from #1246.

Transform runtimes currently receive directive/export metadata but cannot tell whether an inline function captures bound values or inspect the declared parameter shape without reparsing source.

This adds `hasBoundArgs` and `{ count, hasRest }` metadata for inline functions, plus parameter metadata for named, variable, local re-export, and default module exports. Focused tests cover closure captures and all supported export forms.

### DIFF
--- a/packages/plugin-rsc/src/transforms/hoist.test.ts
+++ b/packages/plugin-rsc/src/transforms/hoist.test.ts
@@ -466,11 +466,11 @@ export async function kv() {
       }),
     ).toMatchInlineSnapshot(`
       "
-      export const none = /* #__PURE__ */ $$register($$hoist_0_none, "<id>", "$$hoist_0_none", {"directiveMatch":["use cache",null]});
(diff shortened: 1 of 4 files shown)


hi-ogawa:
The "runtime meta" now provides corresponding function ast node, so this should allow custom transform to make use of it instead of specific metadata `hasBoundArgs/count/hasRest`. Please check out the example here:

https://github.com/vitejs/vite-plugin-react/blob/4b0e3c720ab5877ac6eb8aeb1f59f69a0bb50e43/packages/plugin-rsc/examples/use-cache-callable/callable-cache-plugin.ts#L44-L46

https://github.com/vitejs/vite-plugin-react/blob/4b0e3c720ab5877ac6eb8aeb1f59f69a0bb50e43/packages/plugin-rsc/examples/use-cache-callable/callable-cache-plugin.ts#L125-L139

Please let me know if something still requires from transform utility side.

PR #1259, opened 2026-06-18
feat(rsc): generalize server action transforms

james-elicx:
Extracted from #1246.

`transformServerActionServer` is currently hard-coded to the `"use server"` directive and a single runtime/validation policy. Framework directives otherwise need to duplicate the module-vs-inline dispatch logic.

This adds configurable inline/module directives, runtimes, filtering, validation, directive preservation, and built-in module detection. Direct tests cover each dispatch option.

### DIFF
--- a/packages/plugin-rsc/src/transforms/server-action.test.ts
+++ b/packages/plugin-rsc/src/transforms/server-action.test.ts
@@ -1,0 +1,71 @@
+import { parseAstAsync } from 'vite'
+import { describe, expect, it } from 'vitest'
+import { transformServerActionServer } from './server-action'
+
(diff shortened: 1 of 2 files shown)


hi-ogawa:
This is interesting, but this direction of "unification" isn't likely a good way to settle yet. Ideally, we want to consolidate inine/file directive handling to solve issues such as:
- https://github.com/vitejs/vite-plugin-react/issues/1372
- https://github.com/vitejs/vite-plugin-react/issues/1394

Please feel free to share any finding to solve above. Wrapping two now as `transformServerActionServer` is premature to decide right now, so let me close this.

PR #1260, opened 2026-06-18
feat(rsc): validate inline action lexical expressions

james-elicx:
Extracted from #1246.

Hoisting a directive function changes the meaning of lexical `this`, `super`, and `arguments`. Custom directives that require semantic preservation need a transform-level way to reject these expressions.

This adds opt-in validation scoped to the transformed function, skips nested function scopes, and allows the JSX dev helper source-metadata `this` argument. Tests cover every rejected expression and both exceptions.


hi-ogawa:
This isn't resolved yet. Now tracked in https://github.com/vitejs/vite-plugin-react/issues/1335.

PR #1261, opened 2026-06-18
feat(rsc): export wrapped inline action hoists

james-elicx:
Extracted from #1246.

Inline transforms currently export the raw hoisted implementation. Consumers that need to import the transformed reference can therefore bypass the runtime wrapper.

This adds an opt-in mode that keeps the implementation private and exports the wrapped reference under the hoist name. Tests cover plain actions and closure-bound actions.


hi-ogawa:
Likely the same feature is now covered by `hoistRuntime: true` option. https://github.com/vitejs/vite-plugin-react/blob/4b0e3c720ab5877ac6eb8aeb1f59f69a0bb50e43/packages/plugin-rsc/examples/use-cache-callable/callable-cache-plugin.ts#L56-L60

PR #1264, opened 2026-06-21
fix(rsc): lower client reference preload priority

TheAlexLichter:
This is a PoC for fetch priority of module preloads. It marks additional SSR-discovered Client Component module preloads as low priority. For the client entry and its static dependencies, the normal Vite fetch priority will be preserved.

This should have impact on all SSR React RSC apps (using the latest compatible React and `@vitejs/plugin-rsc`).
They would automatically get a better LCP by reducing Client Component JavaScript competition with assets like critical images.
Client-only RSC apps would not benefit because this behavior occurs during SSR preload generation.

Note: This requires ~~https://github.com/react/react/issues/36834 to be resolved~~ https://github.com/react/react/pull/36835 to be released before this has any effect. I have tested the changes with this POC and it works as expected.


kurtextrem:
What's the thinking behind preloading with lower priority? (asking so I understand the intention)

because I see two things:
1. needed right away in hydration / interactivity, or 
2. not needed urgently

In case 1., lowering prio of _some_ modules does not have a that big impact for LCP, so you may wonder: why not reduce the prio of all modules (preloads)? ... but you can then ask: what is the purpose? even a lower prio preload is still `high` (e.g. this: `<link rel="modulepreload" fetchpriority="low" href="...">` -> still prio `high`). 
The only advantage with `fetchpriority=low` is, Lighthouse might score higher (see https://github.com/GoogleChrome/lighthouse/issues/16539) - but that only impacts *simulated* LCP, not real users.

In case 2.: I'd say we (you) should think of a way to lazily preload the modules (dynamically inject via `requestIdleCallback`? a few seconds after `load`? but how do they get transported to the client? manifest? inline js? etc.)

TheAlexLichter:
Right now, every `modulepreload` is fetched at `High` priority. 

This lets client-reference chunks compete equally with the client entry and render-critical resources. Although these
chunks are required to hydrate their respective client components, the client entry must load first to start hydration.

This PR keeps client-entry dependencies at their default priority (`High` in Chromium), but marks additional client-reference chunks as `Low`. 

They are still discovered and fetched early, avoiding a later fetch-on-import waterfall, while allowing the browser to prioritize CSS, potential LCP resources, and the code that starts hydration when bandwidth is constrained.

---

> Even a lower-priority preload is still high.

This is a weird one!

The [HTML specification](https://html.spec.whatwg.org/multipage/links.html#link-type-modulepreload) says that `modulepreload` should respect `fetchpriority`. Firefox and Safari do so.
Chromium does too, but **only for static/parser-discovered links**, not links dynamically inserted with JavaScript.

Fortunately, the RSC plugin emits these links while streaming the HTML, so they are discovered by the parser and the requested priority is applied.

[Here is a demo](https://modulepreload-fetchpriority-demo.thealexlichter.workers.dev/dynamic) for anyone interested.

<details>
<summary>Demo instructions</summary>
<p>

1. Open Chrome DevTools and select **Network**.
2. Enable **Disable cache** and the **Priority** column.
3. Reload the dynamic page.
4. Find `module.js?dynamic`.

It shows `High` priority despite being created with `fetchPriority = "low"`. Follow the **static control** link and repeat the steps; that request shows `Low`.

</p>
</details>

**TL;DR: The priority hint is still effective in our case.** I also filed a [Chromium bug](https://issues.chromium.org/issues/526039706) for dynamically inserted modulepreloads.

---

The second use case you mentioned is out of scope. This PR only concerns modules required by the initial rendered tree and adjusts their scheduling (according to their role in hydration).

Modules that are not needed for initial hydration should not be preloaded here at all.
Lazy or intent-based preloading would require a separate design (but is a good idea to consider.

kurtextrem:
But why is the client entry then still `high`? What does the client entry do while the other chunks are potentially still loading?

Great find on the bug btw!

TheAlexLichter:
> But why is the client entry then still `high`? What does the client entry do while the other chunks are potentially still loading?

Without the client entry, we cannot do much with the other chunks. The entry initializes React, starts processing the RSC payload, and begins hydrating boundaries (for those dependencies that are already available).

Individual client-reference chunks, on the other hand, only unblock hydration for the components that depend on them.

In turn we want the entry "as early as possible" so the client runtime and hydration can start promptly, while lowering the priority of additional chunks to prevent them from "equally competing" with render-critical resources.

Of course, the amount of useful work possible before those chunks arrive depends on the application and its component/chunk boundaries.

> Great find on the bug btw!

Thanks!

TheAlexLichter:
Moving this into non-draft as the PR got merged (only waiting for release now) but the review can still happen.

hi-ogawa:
Thanks for the upstream fix 👍 
The code looks mostly fine but too AI especially tests, so I may tweak later.
The next canary (today or tomorrow) should probably include the change so we can keep an eye on https://npmx.dev/package/react/versions

TheAlexLichter:
Yes please tweak as needed 👍

hi-ogawa (approved):
Thanks for the fix.
I've updated e2e to actually check fetchpriority on canary and experimental.

TheAlexLichter:
Now we just need to wait for a React release 🔥

PR #1270, opened 2026-06-26
fix(react): babel-plugin-react-compiler cannot be imported when used in a framework

smeng9:
### Description

I am a meta-framework developer that my library incorporates vite's javascript api.
My library is a cli based tool, calls vite's build() function internally, and processes user files.
It is supposed to be able to run regardless of whether there is a local node_module folder in the user's cwd.

I encountered the following issue issue 

@rolldown/plugin-babel calls babel.loadOptionsAsync with { babelrc: false, configFile: false }. When there is no config file, Babel uses process.cwd() as the dirname for resolving plugins in programmatic presets — not the directory of the file being transformed.

reactCompilerPreset at @vitejs/plugin-react/dist/index.js:46 returns:
preset: () => ({ plugins: [["babel-plugin-react-compiler", options]] })
//                           ^ bare string, not a pre-resolved absolute path

Babel must then locate babel-plugin-react-compiler via:
import.meta.resolve("babel-plugin-react-compiler",
  pathToFileURL(path.join(process.cwd(), "babel-virtual-resolve-base.js")).href)

When process.cwd() is a directory that doesn't have babel-plugin-react-compiler in its node_modules ancestry — e.g., / (the Docker container default) — the resolution fails.

It needs to ensure Babel receive an absolute path, so it never needs to do CWD-relative resolution

### DIFF
--- a/packages/plugin-react/src/reactCompilerPreset.ts
+++ b/packages/plugin-react/src/reactCompilerPreset.ts
@@ -2,7 +2,7 @@ import type {
   ReactCompilerBabelPluginOptions,
   RolldownBabelPreset,
 } from '#optionalTypes'
-import { createRequire } from "module";
+import { createRequire } from 'node:module';
 
 export const defaultCodeFilter =
   /forwardRef|memo|(?:const|let|var|function)\s+(?:[A-Z]|use[A-Z0-9])|(?:[A-Z]|use[A-Z0-9])[^\s:=(){}[\],;]*\s*(?:\(|[:=]\s*(?:function|\())/
--- a/packages/plugin-react/src/reactCompilerPreset.ts
+++ b/packages/plugin-react/src/reactCompilerPreset.ts
@@ -2,15 +2,18 @@ import type {
   ReactCompilerBabelPluginOptions,
   RolldownBabelPreset,
 } from '#optionalTypes'
+import { createRequire } from "module";
 
 export const defaultCodeFilter =
   /forwardRef|memo|(?:const|let|var|function)\s+(?:[A-Z]|use[A-Z0-9])|(?:[A-Z]|use[A-Z0-9])[^\s:=(){}[\],;]*\s*(?:\(|[:=]\s*(?:function|\())/
+
+const require = createRequire(import.meta.url);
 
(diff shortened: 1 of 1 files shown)


smeng9 (commented):
use node:module

ArnaudBarre:
In your setup, if there is no node_modules, where is located the code for the react compiler plugin?

smeng9:
Hi @ArnaudBarre 

My tool is supposed to be installed using `npm install -g`, it will be installed into `npm root -g` location.

That means, in a docker node environment, if I run my tool with the cwd as /, it cannot find the babel-plugin-react-compiler from /, but the babel-plugin-react-compiler is indeed installed at the global node_modules.

ArnaudBarre:
Ok thanks for the clarification I see. @sapphi-red what's your opinion on us doing the resolve of babel plugin instead of Babel? We are doing this for SWC but I looked at the PR dropping Babel and we didn't  for the Babel plugins

sapphi-red:
I think we should have been doing it given that `babel-plugin-react-compiler` is a peer dep of plugin-react.

packages/plugin-react/src/reactCompilerPreset.ts
```diff
@@ -2,15 +2,18 @@ import type {
...
   /forwardRef|memo|(?:const|let|var|function)\s+(?:[A-Z]|use[A-Z0-9])|(?:[A-Z]|use[A-Z0-9])[^\s:=(){}[\],;]*\s*(?:\(|[:=]\s*(?:function|\())/
 
+const require = createRequire(import.meta.url);
+
 export const reactCompilerPreset = (
   options: ReactCompilerBabelPluginOptions = {},
 ): RolldownBabelPreset => ({
   preset: () => ({
-    plugins: [['babel-plugin-react-compiler', options]],
+    plugins: [[require.resolve('babel-plugin-react-compiler'), options]],
```
sapphi-red: I think we can use `import.meta.resolve` instead of `require.resolve`

sapphi-red (approved):

ArnaudBarre (approved):

PR #1280, opened 2026-07-11
fix(plugin-rsc): expose `encodeFormAction` option type and re-export `encodeReply` and `createTemporaryReferenceSet` from ssr

dai-shi:
- Related https://github.com/react/react/pull/27563/


dai-shi:
Can anyone trigger preview please?

hi-ogawa:
One question, there's also `createTemporaryReferenceSet` that normally goes with client `encodeReply`, but should that get re-exported too?

dai-shi:
`createTemporaryReferenceSet` is already exported from `/rsc`. Do you mean from `/ssr`?

dai-shi:
https://github.com/wakujs/waku/pull/2191 is my usage.

hi-ogawa:
> `createTemporaryReferenceSet` is already exported from `/rsc`. Do you mean from `/ssr`?

Yes, I was thinking ssr environment now wants `encodeReply({ temporaryReferences: createTemporaryReferenceSet() })` on its own.

dai-shi:
As far as I understand, in the ssr environment, there's no usage to consume the temporary references, so exporting it might not be very useful.

hi-ogawa:
I've dug in and added e2e with `examples/ssr-encode-form-action` to motivate the use case.

From what I can see, it looks like there's a gap in upstream `encodeFormAction` option's capability. I've seen the introduction in https://github.com/react/react/pull/27563/ but for action url marking use case, this looks like upstream should be just exposing their `defaultEncodeFormAction` and surrounding implementation, so that framework side can just tweak action url like:

```js
const customEncodeFormAction: EncodeFormActionCallback = (id, bound) => {
  const defaultResult = defaultEncodeFormAction(id, bound)
  return {
    ...defaultResult,
    action: '/?custom-action=1',
  }
}
```

Also, still then, the `useActionState(... permalink)` use case is a different path, so the responsibility of custom url is on end users. All in all, it looks like upstream API needs polish. 

That said, I'm happy to land the current shape as it just re-exports what upstream allows, but I'm curious if you have an intent to discuss custom `encodeFormAction` API scope with upstream. 

I haven't fully read but I have some write-up of that framing in `packages/plugin-rsc/examples/ssr-encode-form-action/README.md` in case it helps.

dai-shi:
Thanks. Yeah, while I'm not sure how it will become, let me open a new issue in the React repo.

PR #1289, opened 2026-07-16
feat(rsc): ability to not load server reference module during `createFromReadableStream` on `rsc` environment

hi-ogawa:
### Description

- related https://github.com/vitejs/vite-plugin-react/pull/1253

TODO: 
- [x] e2e (payload producer and reviver in different runtime and verify module not loaded on reviver end) 
- [x] docs

This PR adds `createFromReadableStream` options `{ preserveServerReferences: true }` to allow not loading server reference module. The use case is demonstrated in `packages/plugin-rsc/examples/cache-replay` and accompanying e2e. In short, it allows rsc flight payload produced by one runtime to be consumed by other runtime and render it as a form action without importing associated server reference module. 

The implementation idea is a same as `$$decode-client` proxy trick we already use for preserving client reference "entity" without reviving actual "module" during `createFromReadableStream`. This PR added `$$preserve` tag for server reference. This idea is illustrated in https://artifacts.hiro18181.workers.dev/vite-plugin-react-rsc-load-id-journey  (this was created before this PR, so missing `$$preserve` mention).

This is a smaller scope of changes from https://github.com/vitejs/vite-plugin-react/pull/1253 where I can understand the motivation and verify the correctness. The other aspect from the original PR such as framework-owned reference format seems to have issues in entire usages of server action lifecycle (from my reading, the e2e there doesn't support progressive enhancement form action since it only intercepts `loadServerAction`), so it's deferred and will be investigated separately.

For further context that I haven't fully verified in my eye but raised by agents are following:
- Next.js doesn't have a similar mechanism and its cache reply does resolve server action module. So this rsc bundler concept seems specific to rsc plugin.
- Also the dump of local notes as gist https://gist.github.com/hi-ogawa/4d3d734f688c938045a6e199540830b9


packages/plugin-rsc/src/plugin.ts
```diff
@@ -359,9 +359,19 @@ export function vitePluginRscMinimal(
...
-              const meta = Object.values(manager.serverReferenceMetaMap).find(
+              let meta = Object.values(manager.serverReferenceMetaMap).find(
                 (meta) => meta.referenceKey === parsed.id,
               )
+              if (!meta) {
+                try {
+                  await (this.environment as DevEnvironment).transformRequest(
+                    parsed.id,
+                  )
+                } catch {}
```
hi-ogawa, line 381: do we need fs.deny check?

packages/plugin-rsc/examples/rsc-server-reference-replay/src/framework/entry.rsc.tsx
```diff
@@ -0,0 +1,71 @@
...
+  if (request.method === 'POST') {
+    const action = await decodeAction(await request.formData())
+    await action()
+  }
+
+  let bytes: Uint8Array
+  if (url.pathname === '/cache') {
+    const { CachedContent } = await import('../cached-content')
+    const stream = renderToReadableStream(<CachedContent />)
+    bytes = new Uint8Array(await new Response(stream).arrayBuffer())
```
hi-ogawa: slop, not here.

packages/plugin-rsc/examples/rsc-server-reference-replay/src/framework/entry.browser.ts
```diff
@@ -0,0 +1 @@
+// This example only tests progressive enhancement without JavaScript.
```
hi-ogawa: just do boring copy from starter

packages/plugin-rsc/examples/rsc-server-reference-replay/src/action-import-state.ts
```diff
@@ -0,0 +1,9 @@
+let actionImported = false
```
hi-ogawa: maybe just `export const some = { thing: false }`

packages/plugin-rsc/examples/rsc-server-reference-replay/src/action.ts
```diff
@@ -0,0 +1,7 @@
+'use server'
+
+import { markActionImported } from './action-import-state'
+
+markActionImported()
+
+export async function replayedAction() {}
```
hi-ogawa: nit. `function testAction() {}`

packages/plugin-rsc/src/react/rsc/client.ts
```diff
@@ -13,12 +13,20 @@ import type {
 export function createFromReadableStream<T>(
   stream: ReadableStream<Uint8Array>,
   options: CreateFromReadableStreamEdgeOptions = {},
+  extraOptions?: {
+    /**
+     * @experimental
+     */
+    preserveServerReferences?: boolean
+  },
```
hi-ogawa, line 28: TODO: Note that this is disabled default because otherwise `createFromReadableStream` usage for action bind args decoding won't revive server action (or alternatively we would need to opt-out this for this pass):

TODO: link

packages/plugin-rsc/examples/cache-replay/README.md
hi-ogawa, line 1: todo: update

packages/plugin-rsc/src/plugin.ts
```diff
@@ -359,9 +359,19 @@ export function vitePluginRscMinimal(
               }
             }
             if (parsed.type === 'server') {
-              const meta = Object.values(manager.serverReferenceMetaMap).find(
+              let meta = Object.values(manager.serverReferenceMetaMap).find(
                 (meta) => meta.referenceKey === parsed.id,
               )
+              if (!meta) {
+                try {
+                  await (this.environment as DevEnvironment).transformRequest(
```
hi-ogawa: also slop

Issue #1247, opened 2026-06-12
Support Transitive deps in `virtual:vite-rsc/client-references`


fuma-nama:
### Related plugins

- [ ] [plugin-react](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react)

- [ ] [plugin-react-swc](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react-swc)

- [x] [plugin-rsc](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-rsc)

### Describe the bug

In a monorepo, when `pkg-a` imports & uses `pkg-b` in a server component, the `pkg-b` is imported in `virtual:vite-rsc/client-references`.

When the main workspace/app only installs `pkg-a`, it will fail with module resolution error like:

```
Error: [vite+]: Rolldown failed to resolve import "pkg-b" from "virtual:vite-rsc/client-references".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rolldownOptions.external`
```

I checked prior issues like https://github.com/vitejs/vite-plugin-react/issues/826 and https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-rsc/docs/architecture.md, not sure if this is intentional, but I think it's not a design flaw, it's possible to resolve directly to a transitive dep, the error only happens because Vite failed to resolve the transitive dep from workspace.

Currently, it will require a consumer to install all transitive deps of `pkg-a` as long as they'll be included in client-references, but this purely depends on how `pkg-a` uses its transitive deps, consumer has no control/awareness of this. It is adding a lot more complexity to consume RSC-native packages/libraries, I found it very difficult to do without making a custom Vite plugin to workaround this (but very fragile).

I have fixed this with a Vite plugin that simply hook on `resolveId()`, and replace it with the full resolved path of module. Would be great if the RSC plugin can use a similar strategy to prevent issues with transitive deps.

You can see my original fix here: https://github.com/fuma-nama/fumapress/blob/3998fcbd618d8d66faed4154a63fa7d93687c548/packages/core/src/vite.ts#L77

### Reproduction

see https://github.com/fuma-nama/fumapress/issues/39

### Steps to reproduce

Although not a direct illustration, this is the original issue from Fumapress:

1. Open a monorepo.
2. `pnpm create fumapress@0.0.14`, include the created project as a workspace of monorepo.
3. `pnpm build` on the workspace.

[...]

PR #1295, opened 2026-07-20
fix(rsc): pass root importer explicitly instead of `undefined` for `this.resolve` in `rsc:virtual-client-package` 

nihgwu:
### Description

@vitejs/plugin-rsc 0.5.28 added a project-root resolution check for client packages in #1277. In Rolldown builds, passing an undefined importer does not reliably represent the project root and can send importer-sensitive resolvers through a framework-owned dependency graph instead. In our case this caused runaway resolution and module-graph growth, ending in an OOM during the first RSC transform.

Resolve the package from the conventional <root>/index.html importer explicitly. This preserves the #1277 behaviour: packages unavailable at the application root, or resolving to a different copy there, still fall back to the fully resolved module id.

Add a regression guard to the existing transitive client package fixture so build coverage fails if the root probe reaches another resolver without a concrete importer.

Reproduced with identical documentation content:

- @vitejs/plugin-rsc 0.5.27: build succeeds
- 0.5.28: build OOMs
- patched 0.5.28: build succeeds

Related to #1247 and #1277.


nihgwu:
@hi-ogawa can you have a look, latest version blocks our builds 🙏

hi-ogawa:
Doing this explicitly does make sense (I remember just seeing this on Vitest https://github.com/vitest-dev/vitest/blob/86c700ed8b77b7afb0ee9ec677ef8379f34a4bed/packages/vitest/src/node/plugins/vitestResolver.ts#L55) and fine to land nonetheless, but I'd appreciate more content on your setup. 

Vite’s built-in resolver already resolves bare packages from the project root when the importer is undefined, so `undefined` and `<root>/index.html` should behave the same there. The reported difference therefore seems to come from another plugin handling `this.resolve()` first. Could you provide the full Vite config and plugin list from the failing setup so we can understand that interaction?

hi-ogawa (changes requested):
for above

hi-ogawa (approved):

PR #1297, opened 2026-07-21
feat(rsc): initial bundledDev support

### LINKED ISSUE #1217: Full bundle support to vite rsc plugin
### Related plugins

### Description

Hi, I tried to enable `experimental.bundledDev`, kinda working by hacky and scary patch, want to know if there is any plan to support it from plugin-rsc side

### Suggested solution

TBD

### Alternative

_No response_

### Additional context

_No response_

### Validations

nihgwu:
### Description

Fixes #1217.

- Use the RSC browser wrapper as the bundledDev client entry and statically include reachable client references.
- Include server-imported CSS and assets in the client bundle and derive the development assets manifest from bundled output.
- Preserve React refresh and ordinary client HMR while avoiding custom HMR state, timers, cache invalidation, and Vite internal APIs.
- Document that server-graph and client-boundary changes require a dev-server restart until Vite exposes an atomic bundledDev rebuild-and-reload operation for plugins.

This includes the build-only `generateBundle` guard also proposed in #1296 as part of the broader bundledDev integration.

### DIFF
--- a/packages/plugin-rsc/README.md
+++ b/packages/plugin-rsc/README.md
@@ -17,6 +17,13 @@ You can create a starter project by:
 npm create vite@latest -- --template rsc
 ```
 
+### Bundled development
+
+Vite's experimental `bundledDev` mode supports initial RSC rendering and client
+HMR. Changes to the server graph, including client-boundary changes, require a
+dev-server restart. Vite does not yet expose an atomic bundledDev
+rebuild-and-reload operation that plugins can safely request from `watchChange`.
+
 ## Examples
 
 **Start here:** [`./examples/starter`](./examples/starter) - Recommended for understanding the package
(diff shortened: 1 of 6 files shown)


nihgwu:
I tested this PR against a real application with #1295 applied and Vite 8.1.4. The client appeared to hydrate, but the test uncovered a correctness blocker around SSR CSS.

Evidence:

- Normal dev with JavaScript disabled rendered 42 stylesheet links and the complete styled layout.
- bundledDev rendered zero stylesheet links and only one unrelated inline style.
- With JavaScript enabled, route-specific and server-only CSS was still missing.

The underlying problem is that bundledDev transforms client CSS into JavaScript calling `updateStyle`, serves only bundled `memoryFiles`, and does not mount the normal transform middleware. plugin-rsc therefore has no valid transformed stylesheet URL to emit into streamed SSR HTML. Crawling the initial server graph is also insufficient because route-specific CSS can be discovered only during the RSC render.

Restoring source CSS URLs produces 404s. Avoiding the links makes SSR depend on JavaScript and still misses server-only styles. Custom CSS middleware, data URLs, or parsing Vite’s generated JavaScript would reproduce Vite internals and would not be an upstream-ready solution.

This PR should therefore remain blocked and should not close #1217 until Vite exposes a server-side way to obtain servable bundled output URLs for CSS/modules and a safe full-rebuild operation for server-graph changes.

Related Vite core API discussion: https://github.com/vitejs/vite/issues/22991

PR #1309, opened 2026-07-24
fix(rsc): ignore root resolution errors in `rsc:virtual-client-package`

nihgwu:
### Description

The project-root eligibility check for client-package virtualisation added in #1277 assumes resolution returns null when a package subpath is unavailable from the root. It can instead throw when a different version of the same package exists at the root but does not export the requested subpath.

One real dependency shape is parse5 resolving entities/decode and entities/escape from a nested entities@6, while the application root has entities@4. Resolution from the actual importer is valid, but the root probe aborts the build.

Treat root-probe errors like unresolved or mismatched root results so the plugin falls back to the fully resolved module ID. Add an inline E2E fixture with incompatible root and compatible nested versions, covering both development and production builds.

Follow-up to #1277 and #1295.


nihgwu:
@hi-ogawa another blocker for us to upgrade to latest version 🙏

hi-ogawa (approved):
Makes sense to me. Adjusted e2e with better illustration.

PR #1310, opened 2026-07-24
feat(rsc): ability to register server function via plugins

hi-ogawa:
- Related https://github.com/cloudflare/vinext/pull/2156
- Design document https://gist.github.com/hi-ogawa/d92c14b2581e0abf8574cee816f7b3ab

Previously our builtin server function support owned `manager.serverReferenceMetaMap` lifecycle entirely i.e. only `use server` directive can participate in server function logic. This change is aimed to extend the `serverReferenceMetaMap` singleton concept so that user land plugins can manage server reference independently through own transform. The implementation is inspired by how Vinext keeps track of server references https://github.com/cloudflare/vinext/pull/2156 with sandwitched plugins.

See `packages/plugin-rsc/examples/custom-server-function/custom-server-function-plugin.ts` for custom server function plugin example.

As explained in the design document above, this is independent from https://github.com/vitejs/vite-plugin-react/pull/1246 which is for improving __standalone transform utilities__ and not about rsc plugin's builtin server references registration mechanism.

TODO
- [x] review
- [x] e2e
- [ ] doc?

PR #1340, opened 2026-07-28
fix(rsc): preserve server reference metadata compatibility

### LINKED ISSUE #1339: `@vitejs/plugin-rsc` 0.5.31 breaks consumers of `serverReferenceMetaMap`
Was testing out Vinext locally and found out https://github.com/vitejs/vite-plugin-react/issues/1310 broke it.

---

## Description

`@vitejs/plugin-rsc` 0.5.31 replaced `RscPluginManager.serverReferenceMetaMap` with `RscPluginManager.serverReferences.metaMap` as part of the server-reference claims work in [#1310](https://github.com/vitejs/vite-plugin-react/pull/1310).

This breaks the latest published Vinext release during production builds. Vinext `1.0.0-beta.4` declares `@vitejs/plugin-rsc: ^0.5.26`, so a fresh installation can resolve plugin-RSC 0.5.31 even though Vinext still reads `manager.serverReferenceMetaMap`.

## Reproduction

Install these published versions in a minimal Vinext App Router application containing a server action:

```json
{
  "dependencies": {
    "@vitejs/plugin-react": "6.0.4",
    "@vitejs/plugin-rsc": "0.5.31",
    "react": "19.2.7",
    "react-dom": "19.2.7",
    "react-server-dom-webpack": "19.2.7",
    "vinext": "1.0.0-beta.4",
    "vite": "8.1.5"
  }
}
```

Then run:

```sh
pnpm install
pnpm exec vinext build
```

The first two scan builds complete, but the final RSC build fails:

```text
[3/5] build rsc environment...

TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at resolveHasServerActions (.../vinext/dist/index.js:865:17)
```

Pinning plugin-RSC to 0.5.30 makes the same application build successfully with stock Vite 8.1.5.

## Cause

In plugin-RSC 0.5.30, the experimental manager exposes:

```ts
serverReferenceMetaMap: Record<string, ServerReferenceMeta>
```

In 0.5.31, the metadata moved to:

```ts
serverReferences: ServerReferencesManager

// ServerReferencesManager
metaMap: Map<string, ServerReferenceMeta>
```

The new manager supports multiple server-reference claim owners, but the old property is now `undefined`. Existing consumers that call `Object.keys()` or `Object.values()` on it fail at build time.

## Expected Behavior

A package version accepted by Vinext's published plugin-RSC peer range should not crash a production build because the manager metadata property disappeared.

`RscPluginManager` is marked experimental, so changing its shape may be intentional. However, plugin-RSC 0.5.31 is currently installable into the latest published Vinext release without a peer warning, which turns the API migration into a fresh-install runtime failure.

## Possible Resolution

Could plugin-RSC temporarily expose a deprecated read compatibility accessor for `serverReferenceMetaMap`, derived from `serverReferences.metaMap`?

For example:

```ts
/** @deprecated Use serverReferences.metaMap. */
get serverReferenceMetaMap(): Record<string, ServerReferenceMeta> {
  return Object.fromEntries(this.serverReferences.metaMap)
}
```

A getter is preferable to assigning a field alias because `ServerReferencesManager` replaces `metaMap` whenever claims change. The getter always reflects the current derived map.

This would cover Vinext's existing read operations through `Object.keys()`, `Object.values()`, and keyed lookup. It would intentionally not preserve direct mutation of the old record. Vinext only reads the property, so read compatibility is sufficient for the concrete released-package break.

Alternatively, if compatibility should be handled entirely in Vinext, it would help to document the manager migration and coordinate package constraints so incompatible plugin-RSC versions produce an installation warning rather than a build-time `TypeError`.

Related change: [vite-plugin-react PR #1310](https://github.com/vitejs/vite-plugin-react/pull/1310)

hi-ogawa:
- closes https://github.com/vitejs/vite-plugin-react/issues/1339
- follow up to https://github.com/vitejs/vite-plugin-react/pull/1310

This PR restore the deprecated `serverReferenceMetaMap` read API removed in plugin-RSC 0.5.31 by deriving the compatibility record from `serverReferences.metaMap`.

### DIFF
--- a/packages/plugin-rsc/src/plugin.ts
+++ b/packages/plugin-rsc/src/plugin.ts
@@ -127,6 +127,12 @@ class RscPluginManager {
   clientReferenceGroups: Record</* group name*/ string, ClientReferenceMeta[]> =
     {}
   serverReferences: ServerReferencesManager = new ServerReferencesManager(this)
+
+  /** @deprecated Use `serverReferences.metaMap` instead. */
+  get serverReferenceMetaMap(): Record<string, ServerReferenceMeta> {
+    return Object.fromEntries(this.serverReferences.metaMap)
+  }
+
   serverResourcesMetaMap: Record<string, { key: string }> = {}
   environmentImportMetaMap: Record<
     string, // sourceEnv


hi-ogawa:
Manually confirmed that Vinext example works with pkg.pr.new.

PR #1347, opened 2026-07-29
refactor(rsc): use Vite ESTree types

hi-ogawa:
A bit of churns with custom `packages/plugin-rsc/src/transforms/estree.ts` but probably good first step. 

Summary:
- consolidate ast types used by rsc transforms against Vite 8’s exported `ESTree` namespace instead of `@types/estree`.
- we still rely on `estree-walker` but uses it by wrapping with Vite AST types.

TODO:
- can we replace deprecated `parseAstAsync`? (need to check if Vite 7 exports)
- should be able to remove `as Parameters<typeof transformXxx>` like https://github.com/vitejs/vite-plugin-react/blob/b8c9264fe3bec509f64297e14fc0780417acf697/packages/plugin-rsc/examples/custom-server-function/custom-server-function-plugin.ts#L34-L36

### DIFF
--- a/packages/plugin-rsc/src/transforms/estree.ts
+++ b/packages/plugin-rsc/src/transforms/estree.ts
@@ -1,9 +1,48 @@
-import type {} from 'estree'
+import { walk as walkEstree } from 'estree-walker'
+import type { ESTree } from 'vite'
 
-// rollup ast has node position
-declare module 'estree' {
-  interface BaseNode {
-    start: number
-    end: number
-  }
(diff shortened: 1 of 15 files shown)


packages/plugin-rsc/src/transforms/estree.ts
hi-ogawa, line 1: still need to ask agent why this is legit with comments

hi-ogawa:
closing in favor of:
- https://github.com/vitejs/vite-plugin-react/pull/1351

Issue #1350, opened 2026-07-29
Rework `transformWrapExport` and `transformHoistInlineDirective`


hi-ogawa:
- related? https://github.com/vitejs/vite-plugin-react/issues/1336

## Motivation

`transformServerActionServer` currently selects between two distinct transform paths:

https://github.com/vitejs/vite-plugin-react/blob/b8c9264fe3bec509f64297e14fc0780417acf697/packages/plugin-rsc/src/transforms/server-action.ts#L30-L39

The code contains an existing TODO to generalize `transformHoistInlineDirective` for top-level directive cases.

The current split is understandable from the implementation history, but it creates cognitive load because the two paths use different output models:

```text
module-level
  exported implementation
    -> reassign export to runtime(implementation)

inline
  private or generated implementation
    -> runtime(implementation)
    -> generated reference export
    -> optional lexical capture binding
```

Understanding or changing Server Function lowering currently requires keeping both models in mind. The module path also has declaration-specific machinery such as converting exported `const` bindings to `let` so they can be reassigned, while the inline path naturally creates a separate canonical binding.

The goal is to investigate whether internal module-export consumers can adopt the inline path's clearer callable model and share its canonical-binding logic. The inline transform is a comparison point and potential shared-primitive consumer, not a behavioral target of the refactor.

## Three lowering shapes

Assume a simple exported function with no closure captures:

```js
export async function action(value) {
  "use server"
  return value
}
```

### `transformWrapExport`

The current module-level transform:

1. Removes the original export syntax while keeping the local declaration.
2. Reassigns the local binding to `runtime(binding)`.
3. Exports the reassigned binding under the original name.

```js
async function action(value) {
  return value
}

action = runtime(action, 'action')
export { action }
```

An exported `const` must first become `let` so this reassignment is legal.

### Hoist without `hoistRuntime`

The inline transform:

1. Moves the original function body into a generated module-level function.
2. Replaces the original declaration with a binding initialized to `runtime(generatedFunction)`.

```js
export async function $$hoist_action(value) {
  return value
}

export const action = runtime($$hoist_action, '$$hoist_action')
```

The generated function is the reported reference export. The original source name holds the runtime result.

This is behaviorally fine when `runtime(...)` annotates and returns the same function, as built-in `"use server"` does. If `runtime(...)` creates a distinct callable, the reported reference still resolves to the raw generated implementation.

### Hoist with `hoistRuntime`

[...]

PR #1354, opened 2026-07-30
chore(rsc): remove stale source map todo in `transformServerActionServer`

hi-ogawa:
As part of digging into https://github.com/vitejs/vite-plugin-react/issues/1350, it appears this TODO comment has been stale for a long time.

## Summary

Remove the stale source-map TODO above `transformServerActionServer`. The source-map behavior it requested was implemented and subsequently verified before `@vitejs/plugin-rsc` moved into this repository.

## Context

The TODO was originally added in [`hi-ogawa/vite-plugins@4593b255`](https://github.com/hi-ogawa/vite-plugins/commit/4593b255496339da96acbb0193584e309bcb1f13) as part of [hi-ogawa/vite-plugins#797](https://github.com/hi-ogawa/vite-plugins/pull/797). At that point, issue [hi-ogawa/vite-plugins#780](https://github.com/hi-ogawa/vite-plugins/issues/780) tracked that the unwrapped `registerServerReference` call did not yet preserve its original source position.

The immediately following [hi-ogawa/vite-plugins#798](https://github.com/hi-ogawa/vite-plugins/pull/798) implemented the current `MagicString.update(...).move(...)` lowering in `transformWrapExport`. Its adjacent comment documents that generated registration code maps to the original `export` token. Issue #780 then recorded that work as complete.

The user-facing behavior was verified in [hi-ogawa/vite-plugins#781](https://github.com/hi-ogawa/vite-plugins/issues/781), with the remaining Vite module-runner line offset fixed by [hi-ogawa/vite-plugins#810](https://github.com/hi-ogawa/vite-plugins/pull/810). This supports the React development behavior described in [facebook/react#30741](https://github.com/facebook/react/pull/30741), where the `registerServerReference` callsite identifies the server function source location.

The already-stale TODO was later copied into this repository with [vitejs/vite-plugin-react#521](https://github.com/vitejs/vite-plugin-react/pull/521). Removing it does not change transformation or source-map behavior.

## Test

- `pnpm test --run src/transforms/server-action.test.ts src/transforms/wrap-export.test.ts`

### DIFF
--- a/packages/plugin-rsc/src/transforms/server-action.ts
+++ b/packages/plugin-rsc/src/transforms/server-action.ts
@@ -4,9 +4,6 @@ import { transformHoistInlineDirective } from './hoist'
 import { hasDirective } from './utils'
 import { transformWrapExport } from './wrap-export'
 
-// TODO
-// source map for `options.runtime` (registerServerReference) call
-// needs to match original position.
 export function transformServerActionServer(
   input: string,
   ast: ESTree.Program,


hi-ogawa:
Hm, probably this TODO is still valid. It's likely that the referenced PR and use-facing behavior is about `use server` module directive transform from client, so that's about `transformDirectiveProxyExport` and not `transformServerActionServer`. The runtime source map not preserving `transformServerActionServer` probably had zero effect since the original function is the one still appearing in call stack and `runtime` wrapper is just for internal tagging and not function decoration.

PR #1358, opened 2026-07-30
test(rsc): add e2e for react source map features

### LINKED ISSUE #1356: Verify development source locations for transformed Server Functions
- related https://github.com/vitejs/vite-plugin-react/pull/1354

The TODO I've prematurely closed seems to have some concrete consequence and actually more so relevant in https://github.com/vitejs/vite-plugin-react/issues/1350. Tracking it as a new issue here so not to forget.

---

## Motivation

In development, React's server-side `registerServerReference` records an `Error` as the reference's location. When the Server Reference is serialized, React parses the caller frame and sends that location to the browser. The browser uses it to construct a source-located fake Server Function for stacks and developer tooling.

This means the generated `registerServerReference(...)` callsite must map to the original Server Function export.

The current direct-declaration path in `transformWrapExport` deliberately maps generated registration code to the original `export` token. This behavior was implemented by [hi-ogawa/vite-plugins#798](https://github.com/hi-ogawa/vite-plugins/pull/798) and manually verified for direct named exports in [hi-ogawa/vite-plugins#781](https://github.com/hi-ogawa/vite-plugins/issues/781).

Other export forms do not have focused verification. This is also relevant to [#1350](https://github.com/vitejs/vite-plugin-react/issues/1350), because replacing reassignment lowering with a private implementation and generated canonical export changes the generated registration call's position.

## Proposed verification

Add a development fixture where module-level Server Functions cross the RSC boundary into a Client Component, for example as a form action:

```tsx
// actions.ts
'use server'

export default async function save() {}
```

```tsx
// page.tsx
import save from './actions'
import Client from './client'

export default function Page() {
  return <Client action={save} />
}
```

```tsx
// client.tsx
'use client'

export default function Client({ action }) {
  return (
    <form action={action}>
      <button>Save</button>
    </form>
  )
}
```

Cover these module export forms where supported:

- Direct named function export as the known-working baseline.
- Default function export.
- Local export specifier and alias.
- Re-exported Server Function, where mapping to the re-export statement is an acceptable approximation.

In development, the source location exposed for the client-side Server Function should resolve to the corresponding original source export rather than generated registration or trailing export code.

Investigate the most stable automated assertion. Possible approaches include inspecting serialized development Server Reference location metadata or observing a browser stack produced through the source-located fake function. Retain a documented manual React DevTools "Go to Definition" check if the behavior is not exposed through a stable test API.

## Context

- [React PR #30741](https://github.com/facebook/react/pull/30741) documents that the server location comes from the `registerServerReference()` callsite.
- [PR #1354](https://github.com/vitejs/vite-plugin-react/pull/1354) identified that the existing source-map TODO had ambiguous historical coverage.
- This issue is about preserving developer-facing source locations, not preserving the exact output semantics of `transformWrapExport`.

hi-ogawa:
- closes https://github.com/vitejs/vite-plugin-react/issues/1356

This PR adds a dedicated `examples/source-map` e2e for React development source-location features. There were some manual test cases in `examples/basic` and those are ported over to new fixture. Furthermore, many server reference variants are added.

This PR only prepares the infra so we can verify source map fix and also catch regression of currect one in the future.


packages/plugin-rsc/e2e/source-map.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    references: [
+      {
+        // Currently resolves to the rendered section instead of the function.
+        name: 'inline-directive',
+        originalSource: '/src/features/inline-directive/server.tsx',
+        originalLine: 9,
+        originalColumn: 4,
+      },
+    ],
+  },
```
hi-ogawa, line 113: This looks like a regression. Inline use server was working in some old demo like
- https://github.com/hi-ogawa/vite-plugins/issues/781#issuecomment-2849009525

hi-ogawa: It may seem possible that this is Vite 7 vs Vite 8 difference. For now, don't dig deeper and follow-up in https://github.com/vitejs/vite-plugin-react/issues/1361

PR #1370, opened 2026-07-31
refactor(rsc): add `transformModuleExportWrap` and deprecate `transformWrapExport`

hi-ogawa:
- part of https://github.com/vitejs/vite-plugin-react/issues/1350

TODO
- [x] migrate internal usages of `transformWrapExport`
- [x] deprecate `transformWrapExport`
- [ ] (follow-up) unify `transformModuleExportWrap` and `transformHoistInlineDirective` for `transformServerActionServer`

Issue #1336, opened 2026-07-28
Support Framework-Owned `"use cache"` Server Functions


hi-ogawa:
This is the issue to track individual conceptually independent improvements involved in
- https://github.com/vitejs/vite-plugin-react/pull/1246

Research and design context is dumped in
- https://github.com/vitejs/vite-plugin-react/pull/1314

## Feature request

Provide the plugin-rsc infrastructure needed for a framework-owned directive such as `"use cache"` to produce React Server Functions with Next.js-compatible callable and transport behavior, without teaching plugin-rsc the cache implementation details.

## Architecture Boundary

plugin-rsc should own reusable Server Function infrastructure:

- Owner- and environment-scoped server-reference claims (unless proven impossibility of directive composition [#1315](https://github.com/vitejs/vite-plugin-react/pull/1315))
- Module and inline directive transform primitives.
- Closure variable capture encoding/decoding transport.
- Browser and SSR proxy generation.
- Development resolution and production manifest publication.

The framework integration should own `"use cache"` semantics:

- Directive spelling and cache kind selection.
- Cache function wrapper construction and implementation
- Whether a cached callable participates in Server Function transport.

The feature is therefore not a request to hardcode `"use cache"` into plugin-rsc (yet). It is a request to make the proven Server Function extension path sufficient and maintainable for a framework to implement it.

## Work plan

Each work item should include its own development and production E2E that demonstrates the next level of the integration, as PR #1310 and PR #1330 do.

| Work item | Tracking |
| --- | --- |
| Owner-aware server-reference claims and canonical identity | [PR #1310](https://github.com/vitejs/vite-plugin-react/pull/1310) |
| Canonical module-level wrapper for inline directives | [PR #1330](https://github.com/vitejs/vite-plugin-react/pull/1330) |
| Cache-relevant transform metadata | [PR #1246](https://github.com/vitejs/vite-plugin-react/pull/1246) |
| Mixed module/inline directive ownership | Follow-up to [PR #1315](https://github.com/vitejs/vite-plugin-react/pull/1315) |
| Higher-level Server Function helper | Follow-up derived from the completed integration |

### Owner-aware server-reference claims and canonical identity

- ref https://github.com/vitejs/vite-plugin-react/pull/1310

[...]

PR #1380, opened 2026-08-03
fix(rsc): resolve local function exports

hi-ogawa:
- Stacked on #1370
- Part of https://github.com/vitejs/vite-plugin-react/issues/1336

Realized a slight gap with Next.js `use cache` transform wrapping strategy in https://github.com/vitejs/vite-plugin-react/pull/1370. This PR adds a few more logic to align for this specific scenario of accessing local export within the same module.

Interesting, at the same time, it also reveals that, if we ignore such local export access semantics, we can actually simplify https://github.com/vitejs/vite-plugin-react/pull/1370 without hoist technique, which may be what I land first and leave this PR's complete-but-more-complex variant for future iteration.


hi-ogawa:
closing in favor of 
- https://github.com/vitejs/vite-plugin-react/pull/1387

PR #1381, opened 2026-08-03
feat(rsc): add `transformModuleExportWrap` and deprecate `transformWrapExport` (take 2)

hi-ogawa:
- alternative of #1370

This is an alternative direction of https://github.com/vitejs/vite-plugin-react/pull/1380.

`transformModuleExportWrap` has slightly simpler model than `transformWrapExport` where it doesn't try to reassign wrapped export. For example,

```js
// input
'use cache'
export async function someFn() {}

// output: transformModuleExportWrap
async function someFn() {}
const $$module_binding_someFn = __WRAP__(someFn, ...)
export { $$module_binding_someFn as someFn }

// output: transformWrapExport
async function someFn() {}
someFn = __WRAP__(someFn, ...)  // 👈  local access of `someFn` has effect of __WRAP__
export { someFn }
```

This can technically differ how local export is perceived in above case, but this wasn't always the case, for example:

```js
// input
'use cache'
const someFn = async () => {}
export { someFn }

// output: transformModuleExportWrap
const someFn = async () => {}
const $$module_binding_someFn = __WRAP__(someFn, ...)
export { $$module_binding_someFn as someFn }

// output: transformWrapExport
async function someFn() {}
const $$wrap_someFn = __WRAP__(someFn, ...)
export { $$wrap_someFn as someFn };
```

So, `transformModuleExportWrap` in general aligns the transform strategy to a simplified model. If this specific behavior is aimed to align with Next.js `use cache`, then we should aim for larger strategy rework as prototyped in https://github.com/vitejs/vite-plugin-react/pull/1380.


hi-ogawa:
closing in favor of 
- https://github.com/vitejs/vite-plugin-react/pull/1387

PR #1385, opened 2026-08-04
feat(rsc): allow `transformWrapExport` filter by node ast

hi-ogawa:
- extracted from #1381

This PR adds `ModuleExportMeta.valueNode`, so `transformWrapExport` filter can decide which node to wrap based on node ast. 

This allows handling the exotic case of Next.js `use cache` where it allows array/object export to not be rejected. See `use-cache-callable` example for reference.

The additional source-map fixtures characterize the current `transformWrapExport` and `transformModuleExportEffect` behavior across classes, comments, aliases, recursive references, and local binding preservation.

### DIFF
--- a/packages/plugin-rsc/src/transforms/wrap-export.ts
+++ b/packages/plugin-rsc/src/transforms/wrap-export.ts
@@ -8,46 +8,13 @@ import {
 } from './module-export-scan'
 import { validateNonAsyncFunction } from './utils'
 
-type ExportMeta = {
-  /** The directly exported declaration or value expression, when available. */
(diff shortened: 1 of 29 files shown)


packages/plugin-rsc/src/transforms/wrap-export.ts
```diff
@@ -8,44 +8,13 @@ import {
 } from './module-export-scan'
 import { validateNonAsyncFunction } from './utils'
 
-type ExportMeta = {
```
hi-ogawa, line 11: This is also now consolidated with `ModuleExportMeta`.

PR #1386, opened 2026-08-04
fix(rsc): preserve fully filtered export groups

hi-ogawa:
Follow-up to #1385.

`transformWrapExport` currently rewrites export syntax even when every export in a group is rejected by the filter. Depending on the export form, this can change `const` to `let`, move declarations, introduce synthetic default bindings, or replace re-exports with imports and generated exports despite there being nothing to wrap.

This evaluates each export filter once and leaves fully filtered declarations, variable declarations, specifier groups, re-exports, and default exports unchanged. Mixed groups still retain the existing rewrite behavior for selected exports. Filtered exports are not validated, generated, or reported.

Tests:

- `pnpm -C packages/plugin-rsc test --run`
- `pnpm -C packages/plugin-rsc tsc`

### DIFF
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -10,6 +10,7 @@
   },
   "ignorePatterns": [
     "*.snap.*",
+    "packages/plugin-rsc/src/transforms/fixtures/source-map/wrap-export/default-comments.js",
     "typescript-eslint/",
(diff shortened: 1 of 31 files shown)


hi-ogawa:
superseded by https://github.com/vitejs/vite-plugin-react/pull/1387

PR #1389, opened 2026-08-04
feat(rsc): expose callable argument metadata

### LINKED ISSUE #1392: Exclude Undeclared Caller Arguments From `"use cache"` Cache Keys
- Related https://github.com/vercel/next.js/pull/72506

This issue tracks another minor independent divergence from Next.js use cache semantics which requires information provided by ast transform. 

`useActionState` scenario sounds a bit practical use case, but otherwise, this feature may not crucially affect base `use cache` behavior.

---

### Problem

Callers and frameworks can provide more arguments than a cached function declares. These undeclared arguments should not affect its cache key.

This applies to cached functions selected by either a file-level directive or an inline function directive.

A practical example is `useActionState`, which supplies previous state and form data even when the action declares no parameters:

```js
async function getCachedRandomValue() {
  'use cache'
  return Math.random()
}

await getCachedRandomValue(previousStateA, formDataA)
await getCachedRandomValue(previousStateB, formDataB)
```

Both calls should invoke and key the cached function with an empty argument list, so they address the same cache entry.

### Argument Admission

For a statically known function:

- A fixed signature admits only its declared positional arguments.
- A zero-parameter function admits no caller arguments.
- A rest parameter admits all remaining arguments.
- An unknown signature conservatively admits all arguments.

For example:

```js
async function getProduct(accountId, productId) {
  'use cache'
}

getProduct('account-7', 'product-1', 'framework-extra')
```

The admitted invocation is:

```js
getProduct('account-7', 'product-1')
```

and its cache arguments are:

```js
['account-7', 'product-1']
```

The implementation must receive the same admitted arguments used for the key. Otherwise an omitted argument could affect execution without affecting cache identity.

For an inline cached function, transform-bound closure captures are separate from caller-supplied invocation arguments. Argument admission applies after that boundary is identified:

```text
[protected capture envelope, ...caller arguments]
                    |
                    v
decoded captures + admitted caller arguments
```

For example, a fixed two-parameter inline function retains its decoded captures and the first two caller arguments while excluding later framework-supplied arguments. A rest parameter continues to admit all caller arguments. The capture envelope or equivalent boundary is handled separately by https://github.com/vitejs/vite-plugin-react/issues/1393.

### Transform Requirement

The source transform knows the declared parameter count and whether the function has a rest parameter. This information cannot be recovered reliably from `Function.length` because default parameters truncate its value, rest parameters are omitted, and generated wrappers can obscure the original signature.

The transform can either:

- Expose metadata such as `{ count, hasRest }` to the framework runtime.
- Emit the required argument slicing directly.

Unknown functions and unresolved re-exports should use a pass-all fallback.

The same source parameter shape is sufficient for file-level and inline directives. Inline lowering additionally composes it with the capture boundary from issue #1393. A self-describing protected envelope lets the framework wrapper remove that slot before applying `{ count, hasRest }`. If captures remain directly bound as multiple arguments instead, the wrapper needs their generated bound-slot count or an equivalent adapter. That is a capture-boundary representation choice, not a second parameter-admission primitive.

Next.js emits an empty list, a fixed prefix slice, or all arguments based on the source signature in [`server_actions.rs`](https://github.com/vercel/next.js/blob/153bf8ac5fa00888ef5fbb2b65cac12f0942a44f/crates/next-custom-transforms/src/transforms/server_actions.rs#L3005-L3034).

### Scope

This issue covers declared caller-argument admission only.

It does not cover:

- The closure capture encryption or boundary representation itself, tracked by https://github.com/vitejs/vite-plugin-react/issues/1393. This issue only composes with that boundary for inline functions.
- Analysis of whether a declared parameter is actually used.
- Cache storage, serialization, invalidation, or lifetime.
- Generated reference identity.

Next.js currently retains every declared parameter rather than performing parameter-usage analysis.

### Verification

- Zero-parameter functions ignore caller-supplied arguments.
- Fixed signatures admit only their declared positional prefix.
- Different admitted arguments produce different entries.
- Rest parameters admit all remaining arguments.
- Default and destructured parameters count as declared positional inputs.
- Unknown signatures conservatively admit all arguments.
- File-level and inline directives apply the same caller-argument admission.
- Inline captures remain in cache identity while undeclared caller arguments are excluded.
- The behavior works in development and production.

hi-ogawa:
- part of https://github.com/vitejs/vite-plugin-react/issues/1336
- closes https://github.com/vitejs/vite-plugin-react/issues/1392
- related to https://github.com/vercel/next.js/pull/72506
- closes https://github.com/vitejs/vite-plugin-react/pull/1258 (supersedes)

Framework-owned callable wrappers cannot recover source parameter admission or generated closure capture boundaries from `Function.length`. This matters when framework-supplied arguments should not affect fixed-signature cache keys while declared arguments and decoded captures should.

This PR exposes `{ count, hasRest }` for statically known source functions and `boundArgumentCount` for inline directive lowering. The bound count describes the generated call shape, so direct captures occupy their actual number of slots while encoded captures occupy one. The callable cache example consumes this metadata and demonstrates fixed-prefix admission, rest pass-through, and decoded capture identity in development and production.

### DIFF
--- a/packages/plugin-rsc/e2e/use-cache-callable.test.ts
+++ b/packages/plugin-rsc/e2e/use-cache-callable.test.ts
@@ -13,6 +13,20 @@ test.describe('build', () => {
 })
 
 function defineTests(f: Fixture) {
+  test('argument admission and capture identity', async ({ page }) => {
+    using _errors = expectNoPageError(page)
+    await page.goto(f.url('/argument-admission'))
+
+    await expect(page.getByTestId('fixed-extra-admission')).toHaveText(
+      'excluded',
+    )
+    await expect(page.getByTestId('fixed-declared-admission')).toHaveText(
+      'included',
+    )
+    await expect(page.getByTestId('rest-admission')).toHaveText('included')
+    await expect(page.getByTestId('capture-admission')).toHaveText('included')
+  })
+
   test('inline directive', async ({ page }) => {
     using _errors = expectNoPageError(page)
     await page.goto(f.url())
--- a/packages/plugin-rsc/e2e/use-cache-callable.test.ts
+++ b/packages/plugin-rsc/e2e/use-cache-callable.test.ts
@@ -23,6 +23,9 @@ function defineTests(f: Fixture) {
     await expect(page.getByTestId('fixed-declared-admission')).toHaveText(
       'included',
     )
+    await expect(page.getByTestId('fixed-implementation-arguments')).toHaveText(
+      'preserved',
+    )
     await expect(page.getByTestId('rest-admission')).toHaveText('included')
     await expect(page.getByTestId('capture-admission')).toHaveText('included')
   })
--- a/packages/plugin-rsc/examples/use-cache-callable/callable-cache-plugin.ts
+++ b/packages/plugin-rsc/examples/use-cache-callable/callable-cache-plugin.ts
@@ -4,6 +4,7 @@ import {
   transformDirectiveProxyExport,
   transformHoistInlineDirective,
   transformWrapExport,
+  type FunctionParameters,
 } from '@vitejs/plugin-rsc/transforms'
 import { parseAstAsync, type Plugin } from 'vite'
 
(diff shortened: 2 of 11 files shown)


hi-ogawa:
Closing since this should split to
- https://github.com/vitejs/vite-plugin-react/issues/1392
- https://github.com/vitejs/vite-plugin-react/issues/1393

PR #1397, opened 2026-08-05
feat(rsc): allow proxy export filter by node ast

### LINKED ISSUE #1391: Support ast-node-based filter in `transformDirectiveProxyExport`

We have demo-ed this capability for `use cache` imported in server, but not when imported in client directly. https://github.com/vitejs/vite-plugin-react/blob/fba021fe203b198402f6784b8259db077cbecec1/packages/plugin-rsc/examples/use-cache-callable/src/features/file-directive-from-server/action.ts#L5-L8

We would need similar filtering feature for `transformDirectiveProxyExport` to support this use case.

---

### Remaining primitive

Allow `transformDirectiveProxyExport` to select which module exports become Server Reference proxies using the same static export facts available to the RSC-side wrapper transform.

The gap is concrete for a module-level `"use cache"` file containing ordinary object/array exports alongside a cached function:

```js
'use cache'

export const metadata = { title: 'page' }
export const tags = ['cache']
export async function cached(value) {
  return value
}
```

The callable cache example already has this source shape in [file-directive-from-server/action.ts](../../../../../code/others/vite-plugin-react/packages/plugin-rsc/examples/use-cache-callable/src/features/file-directive-from-server/action.ts). Its RSC branch filters object and array exports before `transformWrapExport` validation at [callable-cache-plugin.ts:37](../../../../../code/others/vite-plugin-react/packages/plugin-rsc/examples/use-cache-callable/callable-cache-plugin.ts#L37-L48), so `metadata` and `tags` remain ordinary values while `cached` is wrapped.

The non-RSC branch has no equivalent filter. `transformDirectiveProxyExport` validates every variable initializer and proxies every discovered export at [packages/plugin-rsc/src/transforms/proxy-export.ts:62](../../../../../code/others/vite-plugin-react/packages/plugin-rsc/src/transforms/proxy-export.ts#L62-L100). With the callable example's `rejectNonAsyncFunction: true`, `metadata` or `tags` throws `unsupported non async function` before `cached` can become a proxy. Existing E2E coverage avoids this path: the mixed-export fixture is imported only through the server, while the client-imported fixture exports only a function.

The required primitive is a metadata-aware proxy filter or equivalent selected-export contract. For framework-owned `"use cache"` behavior:

- Known function exports become proxies.
- Unknown expressions remain conservatively proxied because their runtime value is unavailable in the client/SSR graph.
- Known object and array exports are omitted from the proxy module.
- Direct primitive exports may still be rejected according to directive policy.
- Returned `exportNames` and Server Reference claims include only emitted proxies.

This is narrower than server-side classification. The RSC side can already implement `typeof value === 'function' ? registerCacheWrapper(value) : value` through `transformWrapExport` without re-evaluating the source initializer.

### Next.js usage and rationale

[Next.js client fixture 17](https://github.com/vercel/next.js/tree/aae4179ac628e55483b62cd023a7e1827dcef122/crates/next-custom-transforms/tests/fixture/server-actions/client-graph/17) contains a module-level `"use cache"` file exporting an object and array. Its checked client output is empty, proving that known non-function exports do not become Server Reference proxies.

[Client fixture 13](https://github.com/vercel/next.js/tree/aae4179ac628e55483b62cd023a7e1827dcef122/crates/next-custom-transforms/tests/fixture/server-actions/client-graph/13) proxies known functions and statically unknown expressions from a `"use cache"` module. The corresponding classifier excludes object, array, and literal expressions from cache runtime wrapping at [server_actions.rs:2966](https://github.com/vercel/next.js/blob/aae4179ac628e55483b62cd023a7e1827dcef122/crates/next-custom-transforms/src/transforms/server_actions.rs#L2966-L2977).

No single maintained Next.js fixture combines known non-functions and functions in one declaration. Fixtures 13 and 17 plus the shared classifier establish the mixed result: emit proxies for callable or potentially callable exports and omit known object/array exports.

hi-ogawa:
- closes https://github.com/vitejs/vite-plugin-react/issues/1391
- follow-up to https://github.com/vitejs/vite-plugin-react/pull/1385

This is `transformDirectiveProxyExport` version of https://github.com/vitejs/vite-plugin-react/pull/1385 to handle cases where `use cache` with object/array literal imported directly from client component. Again this is arguably next.js quirks, but this can be naturally done by pure refactoring of `transformDirectiveProxyExport` to switch to use the same `scanModuleExports` infra.


packages/plugin-rsc/src/transforms/wrap-export.test.ts
```diff
@@ -369,8 +369,8 @@ export default Page;
   test('filter value node', async () => {
     const input = `\
 export const action = async () => {}
-export const metadata = {}
-export const tags = []
+export const objectValue = {}
+export const arrayValue = []
```
hi-ogawa, line 373: Minor drive-by naming clarification + alignment.

Issue #1193, opened 2026-04-17
Babel plugins that read AST `loc` produce wrong positions when running after OXC


Artur-:
### Related plugins

- [x] [plugin-react](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react)

- [ ] [plugin-react-swc](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react-swc)

- [ ] [plugin-rsc](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-rsc)

### Describe the bug

When a Babel plugin runs after `@vitejs/plugin-react`'s OXC transform (using `@rolldown/plugin-babel` with `enforce: 'post'` or a custom plugin), the Babel AST `loc` values refer to the OXC-transformed code, not the original source. Any Babel plugin that reads `loc.start.line` / `loc.start.column` to embed source positions in the output will produce incorrect values.

This was not an issue with `@vitejs/plugin-react` v5 where Babel handled both JSX transform and custom plugins in a single pass — `loc` always pointed to the original source.

### Example

We have a Babel plugin that inserts `ComponentName.__debugSourceDefine = { fileName, lineNumber, columnNumber }` after each React function component, using the function body's `loc`:

```js
FunctionDeclaration(path, state) {
  const name = path.node.id.name;
  const loc = path.node.body.loc;  // ← wrong after OXC transform
  // inserts: Foo.__debugSourceDefine = { fileName: "...", lineNumber: loc.start.line, ... }
  path.insertAfter(createDebugInfo(name, state.file.opts.filename, loc));
}
```

With v5 (`react({ babel: { plugins: [myPlugin] } })`), `loc.start.line` referred to the original `.tsx` file. With v6, OXC transforms JSX first, changing line numbers. Babel sees the transformed code and its `loc` values are offset.

### The dilemma — no configuration produces correct results

**With `enforce: 'pre'` (default)** — Babel runs before OXC:
- ✅ `__debugSourceDefine` line numbers are correct (Babel sees original source)
- ❌ OXC's `jsxDEV()` source info is **wrong** — every JSX element in the app gets shifted line numbers because Babel inserted `__debugSourceDefine` blocks (4+ lines each) before OXC processed the file

**With `enforce: 'post'`** — Babel runs after OXC:
- ❌ `__debugSourceDefine` line numbers are **wrong** (Babel AST `loc` refers to OXC-transformed code, not the original source)
[...]

PR #1419, opened 2026-08-09
feat(react): add native React Compiler support

Boshen:
Add a `compiler` option backed by the optional `oxc-transform-react` package.

When enabled, `oxc-transform-react` owns the React Compiler, TypeScript/JSX, and Fast Refresh transforms in one pass, while Rolldown React transforms are disabled. React Compiler and Fast Refresh remain client-only; server environments use the same package for TypeScript/JSX.

Keeping React Compiler in a separate package avoids adding framework-specific compiler size and complexity to Rolldown.


packages/plugin-react/src/index.ts
```diff
@@ -262,11 +269,92 @@ export default function viteReact(opts: Options = {}): Plugin[] {
...
+
+  return plugins
+}
+
+function createReactCompilerPlugin(
+  options: ReactCompilerOptions,
+  include: NonNullable<Options['include']>,
+  exclude: NonNullable<Options['exclude']>,
+): Plugin {
+  const { sourcemap = true, ...compilerOptions } = options
```
hi-ogawa: for build, we can derive this default based on `build.sourcemap` config https://vite.dev/config/build-options#build-sourcemap

packages/plugin-react/src/index.ts
```diff
@@ -262,11 +269,92 @@ export default function viteReact(opts: Options = {}): Plugin[] {
...
+        },
+        code:
+          compilerOptions.compilationMode === 'annotation'
+            ? /['"]use memo['"]/
+            : defaultCodeFilter,
+      },
+      async handler(code, id) {
+        let transform: typeof import('oxc-transform-react').transform
+        try {
+          ;({ transform } = await import('oxc-transform-react'))
```
hi-ogawa: The dynamic import check can move to `async config` hook, so the error gets surfaced earlier.

packages/plugin-react/src/index.ts
```diff
@@ -262,11 +269,111 @@ export default function viteReact(opts: Options = {}): Plugin[] {
...
+            ? /['"]use memo['"]/
+            : defaultCodeFilter,
+      },
+      async handler(code, id) {
+        // The config hook is not called when the plugin is used with Rolldown directly.
+        const { transform } =
+          compiler ?? (await loadCompiler((message) => this.error(message)))
+
+        const result = await transform(id.split('?')[0]!, code, {
+          jsx: 'preserve',
```
ArnaudBarre, line 349: Having this plugin doing this transformation with JSX preserve and then the builtin rolldown doing the jsx transformation will make the jsxDev output have wrong line numbers. This breaks various plugin and browser extensions that are using this to jump from the client to the editor in dev mode.

ArnaudBarre (commented):
I'm wondering if we would not have better stability by always using `oxc-transform-react`. You could then remove react refresh from the builtin oxc-transform, making it a lot more vendor neutral.
That's a trade off, because it would be a bit slower for people not using the compiler, but in my experience the overhead is not visible in real Vite application (and in build mode without compiler, the builtin transform could still be used)

packages/plugin-react/src/index.ts
Boshen: Re both comments:

`oxc-transform-react` is intended to provide or react related transforms.

For `@vitejs/plugin-react`, I want to keep the blast radius minimal so it only uses the react compiler transform from `oxc-transform-react`.

> Having this plugin doing this transformation with JSX preserve and then the builtin rolldown doing the jsx transformation will make the jsxDev output have wrong line numbers. 

Is this a bug that I should fix? Does it happen with babel react compiler as well?

ArnaudBarre: It happens when people configure Babel to only use the react compiler,  but it can be fixed by telling Babel to also [handle jsx transform](https://github.com/vitejs/vite-plugin-react/issues/1193#issuecomment-4276204131), which can't be done with the current setup

packages/plugin-react/package.json
```diff
@@ -59,6 +60,7 @@
   "peerDependencies": {
     "@rolldown/plugin-babel": "^0.1.7 || ^0.2.0",
     "babel-plugin-react-compiler": "^1.0.0",
+    "oxc-transform-react": "^0.144.0",
```
sapphi-red: Whenever oxc bumps its minor version, we have to release plugin-react to solve the peer dep error / warning. We can live with it for a while but it's not ideal.

packages/plugin-react/README.md
```diff
@@ -79,13 +79,48 @@ Under the hood, this simply updates the React Fash Refresh runtime URL from `/@r
...
+
+```js
+react({ compiler: { compilationMode: 'annotation' } })
+```
+
+Source maps are generated by default during development. For production builds, the default follows Vite's [`build.sourcemap`](https://vite.dev/config/build-options.html#build-sourcemap) option. They can be disabled explicitly to improve transform performance:
+
+```js
+react({ compiler: { sourcemap: false } })
+```
```
sapphi-red: I don't think we need this option. If `build.sourcemap` is enabled, then disabling this will break the sourcemap and in that case, users should set `build.sourcemap: false`. `sourcemap: false` + dev can technically work, but for that, I think it should be a `dev.sourcemap` option in Vite rather than in this plugin.

sapphi-red (approved):

Boshen:
Release plan: merge this integration PR, then after today's Oxc release (`v0.145.0`), do a bump then a formal release.

Boshen:
@sapphi-red handing over to you and the team.

packages/plugin-react/package.json
hi-ogawa: > Whenever, oxc bumps its minor version, we have to release plugin-react to solve the peer dep error / warning.

Can you explain how this happens?

packages/plugin-react/README.md
```diff
@@ -79,13 +79,42 @@ Under the hood, this simply updates the React Fash Refresh runtime URL from `/@r
...
+### Babel React Compiler
+
+React Compiler can also be used through Babel with the exported `reactCompilerPreset` helper. This requires [`@rolldown/plugin-babel`](https://npmx.dev/package/@rolldown/plugin-babel), [`babel-plugin-react-compiler`](https://npmx.dev/package/babel-plugin-react-compiler), and [`@babel/core`](https://npmx.dev/package/@babel/core) as peer dependencies:
 
 ```sh
 npm install -D @rolldown/plugin-babel @babel/core babel-plugin-react-compiler
 ```
 
-If you are using TypeScript, you will also need to install [`@types/babel__core`](https://npmx.dev/package/@types/babel__core):
+If you are using TypeScript with Babel, you will also need to install [`@types/babel__core`](https://npmx.dev/package/@types/babel__core):
```
ArnaudBarre:
```suggestion
If you are using TypeScript with Babel 7, you will also need to install [`@types/babel__core`](https://npmx.dev/package/@types/babel__core):
```

packages/plugin-react/src/index.ts
```diff
@@ -82,6 +89,14 @@ export default function viteReact(opts: Options = {}): Plugin[] {
...
     enforce: 'pre',
     config(_userConfig, { command }) {
+      if (opts.compiler) {
+        return {
+          oxc: { jsx: 'preserve' },
+          optimizeDeps: {
+            rolldownOptions: { transform: { jsx: 'preserve' } },
+          },
+        }
+      }
```
ArnaudBarre: I think we should not touch the oxc config here
- If the compiler ran, there is no jsx anymore so it doesn't matter
- If the compiler didn't run because of the include/exclude, the jsx should still be transformed to not ship it to the browser

I think the same idea also apply for the two others compiler `opts.compiler` conditions

Boshen: My intent is for `oxc-transform-react` to own React Compiler, TypeScript/JSX, and Fast Refresh in one pass. Vite/Rolldown’s React transforms are disabled to avoid duplicate transforms.

Can you advice on the correct change you are proposing?

ArnaudBarre: If the file was already transform by `oxc-transform-react` then there is no risk of duplicate work for the JSX transform. For fast refresh, it may indeed induce duplicated work.
It's important to not disable JSX transform all together, because there are some users that are using the include/exclude query to opt out of fast refresh transform but still expect jsx transform (when generating PDF from react component in workers for example)

Here is my suggestion that takes into account both things:

```ts
    config(_userConfig, { command }) {
      const refresh = command === 'serve' && opts.compiler
      if (opts.jsxRuntime === 'classic') {
        return {
          oxc: {
            jsx: { runtime: 'classic', refresh },
            jsxRefreshInclude: refresh ? makeIdFiltersToMatchWithQuery(include) : undefined,
            jsxRefreshExclude: refresh ? makeIdFiltersToMatchWithQuery(exclude) : undefined,
          },
        }
      } else {
        return {
          oxc: {
            jsx: {
              runtime: 'automatic',
              importSource: opts.jsxImportSource,
              refresh,
            },
            jsxRefreshInclude: refresh ? makeIdFiltersToMatchWithQuery(include) : undefined,
            jsxRefreshExclude: refresh ? makeIdFiltersToMatchWithQuery(exclude) : undefined,
          },
          optimizeDeps: {
            rolldownOptions: { transform: { jsx: { runtime: 'automatic' } } },
          },
        }
      }
    },
```

packages/plugin-react/src/index.ts
```diff
@@ -169,7 +186,7 @@ export default function viteReact(opts: Options = {}): Plugin[] {
         command,
         userConfig.server?.hmr,
       )
-      if (skipFastRefresh) {
+      if (skipFastRefresh && !opts.compiler) {
```
ArnaudBarre: If compiler is enabled, this should be false already and I prefer to keep this condition simple instead of optimizing for one config merge in the rare usecase covered by this plugin (hmr disabled by plugin IIRC) 
```suggestion
      if (skipFastRefresh) {
```

packages/plugin-react/src/index.ts
```diff
@@ -132,10 +147,12 @@ export default function viteReact(opts: Options = {}): Plugin[] {
     options(options) {
       if (!runningInVite) {
         options.transform ??= {}
-        options.transform.jsx = {
-          runtime: opts.jsxRuntime,
-          importSource: opts.jsxImportSource,
-        }
+        options.transform.jsx = opts.compiler
+          ? 'preserve'
+          : {
```
```suggestion
        options.transform.jsx = {
```

ArnaudBarre (approved):

packages/plugin-react/package.json
sapphi-red: Oxc releases 0.x.0 now. The next version will be `0.145.0`. `^0.144.0` will not match `0.145.0`, so this range would need to be changed to `^0.144.0 || ^0.145.0` to allow the new version to be used.

hi-ogawa: Ah got it. Forgot about 0. version system.

packages/plugin-react/src/index.ts
```diff
@@ -52,6 +53,12 @@ export interface Options {
    * reactRefreshHost: 'http://localhost:3000'
    */
   reactRefreshHost?: string
+  /**
+   * Enable React Compiler with its default options or configure it.
+   * This requires `oxc-transform-react` to be installed.
+   * @default false
```
sapphi-red, line 59:
```suggestion
   * @default false
   * @experimental
```

PR #1431, opened 2026-08-21
fix(react): make logging diagnostics an opt-in for React Compiler

dqbd:
### Description

When Rust React Compiler is enabled in `@vite/plugin-react@6.1.0`, the plugin emits a Vite warning for every recoverable diagnostic, making the console output unnecessarily verbose. 

This PR moves the diagnostics logging behind a `logDiagnostics` plugin option, with the option being disabled by default.

### DIFF
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -20,6 +20,15 @@ import { defaultCodeFilter, reactCompilerPreset } from './reactCompilerPreset'
 const _dirname = dirname(fileURLToPath(import.meta.url))
 const refreshRuntimePath = join(_dirname, 'refresh-runtime.js')
 
+interface ReactCompilerPluginOptions extends ReactCompilerOptions {
+  /**
+   * Log recoverable React Compiler diagnostics through Vite.
+   * Fatal diagnostics are always logged and fail the transform.
+   * @default false
+   */
+  logDiagnostics?: boolean
+}
+
 export interface Options {
   /**
    * Can be used to process extra files like `.mdx`
@@ -59,7 +68,7 @@ export interface Options {
    * @default false
    * @experimental
    */
-  compiler?: boolean | ReactCompilerOptions
+  compiler?: boolean | ReactCompilerPluginOptions
 }
 
 const defaultIncludeRE = /\.[tj]sx?$/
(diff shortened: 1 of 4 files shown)


packages/plugin-react/src/index.ts
```diff
@@ -288,7 +297,7 @@ export default function viteReact(opts: Options = {}): Plugin[] {
 }
 
 function createReactCompilerPlugin(
-  options: ReactCompilerOptions,
+  options: ReactCompilerPluginOptions,
```
```suggestion
 { logDiagnostics, ...reactCompilerOptions }: ReactCompilerPluginOptions,
```

packages/plugin-react/tests/reactCompiler.test.ts
```diff
@@ -101,12 +101,56 @@ describe('compiler option', () => {
...
+      (diagnostic) => diagnostics.push(diagnostic),
+    )
+
+    expect(diagnostics).toHaveLength(1)
+    expect(diagnostics[0]).toContain(
+      'Hooks must always be called in a consistent order',
+    )
+  })
+
+  test('fails on fatal diagnostics when logging is disabled', async () => {
```
ArnaudBarre: This is just testing the oxc-transform library, not the plugin

dqbd: Removed the test!

packages/plugin-react/src/index.ts
dqbd: Updated!

ArnaudBarre:
Thanks for the update! Can you document this new option on the readme and add a changelog entry?
@sapphi-red an opinion on the default behaviour?

dqbd:
Added, thank you @ArnaudBarre!

ArnaudBarre (approved):

sapphi-red:
> @sapphi-red an opinion on the default behaviour?

Given that the babel plugin (and [Next.js](https://github.com/vercel/next.js/blob/d4b7cc1fb48dce907093b76b6230db3385a10623/turbopack/crates/turbopack-ecmascript/src/transform/mod.rs#L565-L567)) doesn't emit these logs, I think we can skip these by default.

packages/plugin-react/tests/reactCompiler.test.ts
```diff
@@ -101,12 +101,38 @@ describe('compiler option', () => {
     expect(withoutSourcemap.map).toBeFalsy()
     expect((await transformWithBuildConfig({}, true)).map).toBeTruthy()
   })
+
+  test('logs recoverable diagnostics by default', async () => {
```
sapphi-red:
```suggestion
  test('logs recoverable diagnostics when enabled', async () => {
```

PR #1438, opened 2026-08-27
feat(react): use SSR transform of React Compiler for server envs

sapphi-red:
### Description

Set `outputMode=ssr` option in React Compiler for server environments.

This has two problems:

- The `outputMode=ssr` is probably not meant to be public yet. It's not released in babel one (https://github.com/react/react/commit/4cf770d7e1a52c66401b42c7d135f40b7dc23981, https://github.com/react/react/commit/50e7ec8a694072fd6fcd52182df8a75211bf084d)
- `consumer=server` does not necessarily mean SSR. It may be an environment for RSC.

I found the option so I thought to fix it but found the problems above so I'm opening this PR as draft to discuss about these problems.

### DIFF
--- a/packages/plugin-react/tests/reactCompiler.test.ts
+++ b/packages/plugin-react/tests/reactCompiler.test.ts
@@ -88,11 +88,13 @@ describe('compiler option', () => {
     })
   })
 
-  test('uses the native JSX transform for server environments', async () => {
+  test('uses the SSR transform for server environments', async () => {
     const output = await transformWithBuildConfig({}, false, 'server')
 
     expect(output.code).toContain('react/jsx-runtime')
     expect(output.code).not.toContain('react/compiler-runtime')
+    expect(output.code).toContain('const count = 0')
+    expect(output.code).not.toContain('useState(0)')
   })
 
   test('uses the Vite build sourcemap setting', async () => {
@@ -144,9 +146,14 @@ async function transformWithBuildConfig(
   }
   return plugin.transform.handler.call(
     context as any,
-    `export function App({ name }) { return <div>{name}</div> }`,
+    /* jsx */ `import { useState } from 'react';
+
+export function App({ name }) {
+  const [count] = useState(0);
+  return <div>{name}{count}</div>
+}`,
     '/entry.tsx',
-  )
+  ) as { code: string; map: any }
 }
 
 async function getViteReactConfig(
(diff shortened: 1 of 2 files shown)


packages/plugin-react/tests/reactCompiler.test.ts
```diff
@@ -88,11 +88,13 @@ describe('compiler option', () => {
...
   })
 
-  test('uses the native JSX transform for server environments', async () => {
+  test('uses the SSR transform for server environments', async () => {
     const output = await transformWithBuildConfig({}, false, 'server')
 
     expect(output.code).toContain('react/jsx-runtime')
     expect(output.code).not.toContain('react/compiler-runtime')
+    expect(output.code).toContain('const count = 0')
+    expect(output.code).not.toContain('useState(0)')
```
sapphi-red, line 97: The output looks like
```js
import { jsxs as _jsxs } from "react/jsx-runtime";
export function App(t0) {
	const { name } = t0;
	const count = 0;
	return /* @__PURE__ */ _jsxs("div", { children: [name, count] });
}
```

hi-ogawa:
Next has something similar and the one in the react compiler isn't likely directly usable (as in it may not be "production ready" for some edge cases).

The agents dug up some history:

- https://github.com/vercel/next.js/pull/54925 September 2023. Added `optimizeServerReact` to remove `useEffect`/`useLayoutEffect` and simplify `useState` in server bundles.
- https://github.com/vercel/next.js/pull/62658 March 2024. Enabled it by default, but disabled the `useState` optimization because of React 19 compatibility concerns.
- https://github.com/vercel/next.js/pull/69788 September 2024. Temporarily disabled it because removing effects too early also hid referenced Server Actions.
- https://github.com/vercel/next.js/pull/70101 and https://github.com/vercel/next.js/pull/70151 September 2024. Re-enabled it by marking effect calls for later dead-code elimination.

Currently it:

- Removes production-server `useEffect` and `useLayoutEffect` calls when dependencies have no observable side effects.
- Does not optimize `useState`, although that implementation remains behind `optimize_use_state: false`.
- Is enabled by default for production webpack server builds.
- Is not currently wired into Turbopack because its server transform rule is commented out.

hi-ogawa:
> consumer=server does not necessarily mean SSR. It may be an environment for RSC.

In rsc environment, `useState/useEffect/etc` is a runtime error since they are undefined in `react-server` exports, so the transform would ideally surfaces the usage error during build time instead like https://nextjs.org/docs/messages/react-client-hook-in-server-component. Rsc plugin currently doesn't do anything (yet).

packages/plugin-react/tests/reactCompiler.test.ts
ArnaudBarre, line 97:
```suggestion
    expect(output.code).not.toContain('useState')
```

ArnaudBarre (approved):

PR #1449, opened 2026-09-09
feat: pull fixes from React 19.3 to refresh runtime

ArnaudBarre:
Pull [two commits](https://github.com/react/react/commits/main/packages/react-refresh/src/ReactFreshRuntime.js) publish as part of [React 19.3](https://github.com/react/react/releases/tag/v19.3.0)


sapphi-red (approved):

ArnaudBarre:
These are kind of edge cases for HMR which should also depend on the React version you are running, so I don't think making a changelog would help people

PR #1450, opened 2026-09-11
fix(react): respect tsconfig jsxImportSource (fix #1448)

### LINKED ISSUE #1448: react({ compiler: true }) ignores tsconfig jsxImportSource
### Describe the bug

Enabling the native React Compiler with `react({ compiler: true })` changes the JSX runtime selected from each file's nearest `tsconfig.json`. With no explicit plugin-level `jsxImportSource`, a file configured to use `@emotion/react` instead imports `react/jsx-runtime`.

Expected: enabling the compiler should preserve the [documented per-file tsconfig inference](https://github.com/vitejs/vite-plugin-react/tree/plugin-react%406.1.1/packages/plugin-react#jsximportsource).

Two identical TSX files produce these imports:

| Plugin options                                          | `styled/component.tsx` (`jsxImportSource: "@emotion/react"`) | `plain/component.tsx` (default React) |
| ------------------------------------------------------- | ------------------------------------------------------------ | ------------------------------------- |
| `{ compiler: false }`                                   | `@emotion/react/jsx-runtime`                                 | `react/jsx-runtime`                   |
| `{ compiler: true }`                                    | **`react/jsx-runtime` — incorrect**                          | `react/jsx-runtime`                   |
| `{ compiler: true, jsxImportSource: "@emotion/react" }` | `@emotion/react/jsx-runtime`                                 | `@emotion/react/jsx-runtime`          |

The last row is expected global-override behavior, not another bug: setting one global source cannot preserve both per-file configurations.

### Reproduction

Self-contained files below. No pragmas, Babel, CSS plugins, or workspace tooling are involved. Runtime imports are externalized so the test can inspect the emitted specifiers without installing or executing React/Emotion.

Create the files in an empty directory, then run:

```sh
npm install --ignore-scripts
npm run repro
```

The script checks the compiler-disabled baseline, prints all three configurations, and exits with code 1 for the compiler-enabled regression. The final global-override demonstration does not affect the exit code.

<details>
<summary>Complete reproduction files</summary>

**`package.json`**

```json
{
  "name": "vite-jsx-import-source-repro",
  "private": true,
  "type": "module",
  "scripts": {
    "repro": "node repro.mjs"
  },
  "devDependencies": {
    "@vitejs/plugin-react": "6.1.1",
    "oxc-transform-react": "0.145.0",
    "vite": "8.2.2"
  }
}
```

**`styled/tsconfig.json`**

```json
{
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "@emotion/react"
  }
}
```

**`plain/tsconfig.json`**

```json
{
  "compilerOptions": {
    "jsx": "react-jsx"
  }
}
```

**Both `styled/component.tsx` and `plain/component.tsx`:**

```tsx
export function Component() {
  return <div>Hello</div>;
}
```

**`repro.mjs`**

```js
import assert from "node:assert/strict";
import { fileURLToPath } from "node:url";

import react from "@vitejs/plugin-react";
import { build } from "vite";

const root = fileURLToPath(new URL(".", import.meta.url));
const expected = {
  styled: "@emotion/react/jsx-runtime",
  plain: "react/jsx-runtime",
};

let failures = 0;
for (const options of [
  { compiler: false },
  { compiler: true },
  { compiler: true, jsxImportSource: "@emotion/react" },
]) {
  const buildOutput = await build({
    root,
    configFile: false,
    logLevel: "silent",
    plugins: [react(options)],
    build: {
      write: false,
      minify: false,
      lib: {
        entry: {
          styled: `${root}styled/component.tsx`,
          plain: `${root}plain/component.tsx`,
        },
        formats: ["es"],
      },
      rolldownOptions: {
        external: /^(react|@emotion\/react)(\/|$)/,
        output: { preserveModules: true },
      },
    },
  });

  console.log(JSON.stringify(options));
  const chunks = [buildOutput].flat().flatMap((output) => output.output);
  for (const [entry, runtime] of Object.entries(expected)) {
    const chunk = chunks.find(
      (output) =>
        output.type === "chunk" &&
        output.facadeModuleId === `${root}${entry}/component.tsx`,
    );
    assert.ok(chunk, `Missing entry: ${entry}`);
    const imports = chunk.imports.filter((specifier) =>
      specifier.endsWith("/jsx-runtime"),
    );
    const matches = imports.length === 1 && imports[0] === runtime;
    console.log(
      `  ${entry}: ${imports.join(", ")} (${matches ? "PASS" : "FAIL"})`,
    );
    if (!options.compiler) assert.deepEqual(imports, [runtime]);
    if (!matches && !options.jsxImportSource) failures++;
  }
}

process.exitCode = failures ? 1 : 0;
```

</details>

### System Info

```text
macOS 26.4.1 arm64
Node 22.14.0
npm 10.9.2
vite 8.2.2
@vitejs/plugin-react 6.1.1
oxc-transform-react 0.145.0
```

Also reproduced after a clean `npm ci --ignore-scripts` using the generated lockfile.

### Related context

- #726 restored tsconfig inference in the regular JSX transform path. It predates the native compiler integration from #1419 and does not cover this compiler-enabled path.
- #1224 concerns file-level JSX pragma comments being moved below imports. This reproduction contains no pragmas, so moving comments cannot address this case.
- In the native compiler transform, `jsx.importSource` is supplied from `reactOptions.jsxImportSource`, and JSX is already lowered before Vite's regular transform can apply the file's tsconfig.

cpruijsen:
### Description

Fixes #1448.

`react({ compiler: true })` lowers JSX in `vite:react-compiler` with `oxc-transform-react`. That transform defaults `jsx.importSource` to `"react"` when the option is unset. Vite's oxc transform never sees the original JSX, so the per-file tsconfig inference restored in #726 does not apply.

With no plugin-level `jsxImportSource`, the compiler transform now reads `compilerOptions.jsxImportSource` from the file's tsconfig using the same `resolveTsconfig` Vite uses for oxc JSX. An explicit `jsxImportSource` option still overrides every file.

Reproduction from the issue (two TSX files, no pragmas):

| Plugin options | file with `jsxImportSource: "@emotion/react"` | default React file |
| --- | --- | --- |
| `{ compiler: false }` | `@emotion/react/jsx-runtime` | `react/jsx-runtime` |
| `{ compiler: true }` (before) | `react/jsx-runtime` | `react/jsx-runtime` |
| `{ compiler: true }` (after) | `@emotion/react/jsx-runtime` | `react/jsx-runtime` |
| `{ compiler: true, jsxImportSource: "@emotion/react" }` | `@emotion/react/jsx-runtime` | `@emotion/react/jsx-runtime` |

**What I chose:** fill `jsx.importSource` from the file's tsconfig inside the compiler plugin.

**Alternative:** set `jsx: "preserve"` in the compiler plugin so Vite's oxc transform keeps owning JSX, Fast Refresh, and tsconfig inference.

**Why:** #1419 put JSX and Fast Refresh in the compiler plugin so the compiler sees the original AST. This change only supplies the missing import source and stays reversible. Happy to switch to preserve-JSX if that is preferred.

`resolveTsconfig` is a hidden rolldown export currently used by Vite. It is loaded through Vite's installation so this package does not gain a new runtime dependency. If you would rather have a public Vite helper, this call site is the consumer.

### Checklist

- Read the Contributing Guidelines.
- Check that there isn't already a PR that solves the problem the same way.
- Documentation already states that `jsxImportSource` is inferred from tsconfig; no extra docs.
- Tests fail without this change and pass with it (`packages/plugin-react` unit tests).

### DIFF
--- a/packages/plugin-react/CHANGELOG.md
+++ b/packages/plugin-react/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Respect tsconfig `jsxImportSource` when using the native React Compiler ([#1448](https://github.com/vitejs/vite-plugin-react/issues/1448))
+
+Enabling `compiler: true` no longer ignores per-file `jsxImportSource` from the nearest tsconfig. Inference now matches the non-compiler JSX path, and an explicit plugin `jsxImportSource` still overrides it.
+
 ## 6.1.1 (2026-08-28)
 
 ### Add `compiler.logDiagnostics` option
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -304,7 +304,6 @@ interface TsconfigResolution {
   ) => {
     tsconfig: { compilerOptions?: { jsxImportSource?: string } }
   } | null
-  TsconfigCache: new () => { clear(): void }
 }
 
 let tsconfigResolutionPromise:
@@ -337,7 +336,6 @@ function createReactCompilerPlugin(
 ): Plugin {
   let jsxDevelopment = false
   let compiler: typeof import('oxc-transform-react') | undefined
-  let tsconfigCache: { clear(): void } | undefined
   let resolveTsconfig: TsconfigResolution['resolveTsconfig'] | undefined
   const runtime =
     reactCompilerOptions.target === '17' || reactCompilerOptions.target === '18'
@@ -399,11 +397,11 @@ function createReactCompilerPlugin(
               const api = await loadTsconfigResolution()
               if (api) {
                 resolveTsconfig = api.resolveTsconfig
-                tsconfigCache = new api.TsconfigCache()
               }
             }
-            importSource = resolveTsconfig?.(filename, tsconfigCache)?.tsconfig
-              .compilerOptions?.jsxImportSource
+            importSource =
+              resolveTsconfig?.(filename)?.tsconfig.compilerOptions
+                ?.jsxImportSource
           } catch {
             // Keep oxc-transform-react's default (`react`) when tsconfig
             // lookup is unavailable, e.g. raw Rolldown without Vite.
--- a/packages/plugin-react/src/index.ts
+++ b/packages/plugin-react/src/index.ts
@@ -1,6 +1,7 @@
 import { readFileSync } from 'node:fs'
+import { createRequire } from 'node:module'
 import { dirname, join } from 'node:path'
-import { fileURLToPath } from 'node:url'
+import { fileURLToPath, pathToFileURL } from 'node:url'
 import {
   exactRegex,
   makeIdFiltersToMatchWithQuery,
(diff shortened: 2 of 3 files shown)


doctor8296:
The initial per-file jsxImportSource case works, but the new TsconfigCache becomes stale in a running dev server.
Tested 3beee4f with Vite 8.2.2, Rolldown 1.2.6 and oxc-transform-react 0.147.0:
1. Start Vite with react({ compiler: true }), no global override, and tsconfig pointing to runtime A.
2. Change jsxImportSource to runtime B. Vite detects the change and sends full-reload, but subsequent HTTP responses still import A.
3. Edit the component: the response contains the updated code, but still imports A. Removing jsxImportSource also leaves A instead of reverting to React.

With compiler: false, the same test correctly switches A → B → React. Adding only tsconfigCache?.clear() before the new resolveTsconfig call also makes the native-compiler test pass. This isolates the problem to the cache introduced here, rather than stale HTTP/module responses.
Could you invalidate this cache on relevant tsconfig changes and add a dev-server regression test? Clearing it on every transform is a diagnostic control, but not sure if it solves bug correctly.

PR #1461, opened 2026-09-16
feat(react-swc): add compiler option

### LINKED ISSUE #428: React Compiler with plugin-swc
Will there be instructions on how to use the new React Compiler given that it's babel based?

ArnaudBarre:
Fixes #428


sapphi-red:
I think we need to add `|| options.compiler` here, otherwise the compiler won't run in build.
https://github.com/vitejs/vite-plugin-react/blob/16381a7be5973d9b41e6868366f32c31aa2b5610/packages/plugin-react-swc/src/index.ts#L241

sapphi-red (approved):

